PR #1, opened 2023-03-19
Added support for GPT3.5 Turbo model

koganei:
Added support to pick the gpt-3.5-turbo model in the settings since I don't have access to gpt-4 yet.

### DIFF
--- a/components/ChatMessage.tsx
+++ b/components/ChatMessage.tsx
@@ -28,6 +28,7 @@ export interface UiMessage {
   sender: 'You' | 'Bot' | string;
   role: 'assistant' | 'system' | 'user';
   text: string;
+  model: string;
   avatar: string | React.ElementType | null;
 }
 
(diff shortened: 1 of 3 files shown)


peperunas:
This solves part of #2.

koganei:
I've tested it and it works for me. Can someone test that gpt-4 still works well?

peperunas:
It would be good to reflect the name of the model in the page title as well

koganei:
@peperunas Agreed. The title currently just doesn't exist. We could maybe have a larger conversation around what the title should be and whether the project needs a more identifiable name.

peperunas:
I was thinking about having some kind of status indicator on the page as well, indicating the discussion's active model.

Somewhere around the bottom right corner of the page, maybe?

enricoros:
Patch well done! Really needed.

PR #7, opened 2023-03-19
Add ability to set custom system message

typpo:
Quick change that make it possible to set a custom system message.  

https://user-images.githubusercontent.com/310310/226215554-096c4124-e426-41db-8c34-9f7b017b30fb.mp4


enricoros:
Thanks! Important to have this. We can clean up the UI later.

PR #15, opened 2023-03-21
Add Voice Input button to Message Box

koganei:
It uses the browser's SpeechRecognition API

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,9 +25,9 @@ Click to fork and run on Vercel, or check out the demo (bring your own API keys)
 - [x] User-provided API keys
 - [x] 🎉 NEW Support for GPT 3.5 Turbo - @koganei
 - [x] 🎉 NEW Custom system purpose - @typpo
+- [x] Voice input
 - [ ] Developer-oriented features
 - [ ] Stop generation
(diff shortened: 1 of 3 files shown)


koganei:
My next step after this is to use the Speech synthesis to add to each message box from the AI.
And then the step after that is to do a 'hands free' mode to be able to have a conversation.

enricoros:
This looks great! Maybe a function is missing? I'll test the parch locally as well. Such a great feature.

enricoros:
Update: merging it.
 - I'm removing the microphone where 'isSpeechEnabled' = false (browser without support for the Speech API)
 - In turn React gives me a "hydration error"
 - So I'm creating a React hook "useSpeechRecognition()" which should solve this
 Just gonna take a little longer to finish the porting and test.
 Note: I'm actually porting with GPT4 - it knows way more than me.

enricoros:
Merged! Please take a look, the change took a while, but was double-checked (if not authored) by gpt4 itself :)

koganei:
Okay that's great. I thought about making that hook, but I didn't have as
much time as I wanted. Thank you for taking the time to enhancing the
feature!

Issue #35, opened 2023-03-30
Purposes / Templating RFC


enricoros:
Hello everyone,

We're working on enhancing our application by focusing on customization, particularly `System Purposes`. As the project advances, we'll be incorporating more features, such as templating and reasoning systems (ReAct, etc).

We'd appreciate your ideas and feedback on:
- Should we include an in-app editor for Purposes and more?
- How can we ensure persistence for these changes? Are there any API-key based CMS options we should consider?
- Would it be beneficial to allow users to contribute their own Purposes, creating a central repository with usage stats and highlighting top shared Purposes?
- If so, are there any centralized or decentralized services that are easy to access?

We look forward to hearing your thoughts and working together to improve the app's customization capabilities.

romainbessugesmeusy:
Hi there and thanks for this wonderful app.
I'm in favor of allowing the customization of Purposes via a `purposes.json` file, following a public JSON schema that you could define. This schema could then be used by any dev / admin to create their own purposes and manage them in the JSON compatible editor of their choice. 
This file could be referred defined as a URL in the .env file / localstorage like OPENAI_API_KEY

enricoros:
Thanks for taking about this possible workflow @romainbessugesmeusy 

I believe I shall implement the idea above. It's an easy step to go in the right direction. Maybe as fist phase, a single JSON file at build time, then loaded dynamically at runtime, then configurable in the environment and then finally the editing in the UI itself.

After that, I want to think at how to share prompt templates with other people and apps. Community templating should be a thing.

jondwillis:
See also: https://github.com/PhantasticUniverse/CollectiveGenius-AutoGPT-Agent-Hub

tmajibon:
Throwing in a suggestion that down the line Personas could be expanded more into Identities. Obviously a ways out when we can't even save the custom Persona yet.

The thought in my mind is you would be able to set for each:
* A name (even later possibility to potentially reference in a more "group" conversation style)
* A prefered model
* A voice

While this definitely feels like fluff at first, it could be used to more fluidly differentiate what models and configurations you're using (ie. if you have speech turned on by default, the voice will let you know immediately which "purpose" it's using and likely which model). Names could also be used instead of model entries for a "friendlier" appearance (when the model is the default for that identity)

younes-io:
Not sure if this is the right issue to mention this, but you let me know.
[...]

PR #37, opened 2023-03-31
Another Pull Request with style changes to prompting bar with tooltips and Content Mode

willscottrod:
Thanks for your feedback, I've done another pull request with the most recent version.

I see you managed to fix a lot of the issues I was looking at doing and the split componentry is such a huge help!

Cheers
Will

### DIFF
--- a/components/Composer.tsx
+++ b/components/Composer.tsx
@@ -12,7 +12,6 @@ import TelegramIcon from '@mui/icons-material/Telegram';
 import { useComposerStore } from '@/lib/store';
 import { useSpeechRecognition } from '@/lib/use-speech-recognition';
 
-
 /// Text template helpers
 
 const PromptTemplates: { [key: string]: string } = {
@@ -21,13 +20,13 @@ const PromptTemplates: { [key: string]: string } = {
   PasteFile: '{{input}}\n\n```{{fileName}}\n{{fileText}}\n```\n',
 };
 
-const expandPromptTemplate = (template: string, dict: object) => (inputValue: string): string => {
-  let expanded = template.replaceAll('{{input}}', (inputValue || '').trim()).trim();
-  for (const [key, value] of Object.entries(dict))
-    expanded = expanded.replaceAll(`{{${key}}}`, value);
-  return expanded;
-};
-
+const expandPromptTemplate =
+  (template: string, dict: object) =>
+  (inputValue: string): string => {
+    let expanded = template.replaceAll('{{input}}', (inputValue || '').trim()).trim();
+    for (const [key, value] of Object.entries(dict)) expanded = expanded.replaceAll(`{{${key}}}`, value);
+    return expanded;
+  };
 
 /**
  * A React component for composing and sending messages in a chat-like interface.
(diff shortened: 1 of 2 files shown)


enricoros:
Thanks, I've reviewed this new version. 

Did you know your app is live already for anyone to use: [link](https://nextjs-chatgpt-app-git-fork-willscottrod-main-enricoros.vercel.app/)

The quality of this patch has improved already, and glad you're figuring out the workflow to stay up to date on the latest changes. However, a couple of notes:
- somehow the commit adds +131 lines of code and removes −129. this is quite large, and would be overwritten by future code that I write. This is because the code is formatted by "Prettier" (https://prettier.io/) on my end, while (VSCode?) on your end applies different spacing and all. This translates in a lot of code being modified, and the patch being harder to maintain because the 2 systems would keep changing the code back and forth. Is there a "Prettier" plugin you can install on your IDE? After installing, load the rules in the ".prettierrc" file, and the systems will be on the same page
- Change 1. UI: I like the looks actually, but also see the value in the existing UI for developers / people that need more space for the edit box (to drop files in, etc). I'm thinking on whether I should make it switchable, with an ON/OFF "ChatGPT-like UI" option.
- Change 2. Prompt: I'm thinking about how to enable people to publish their prompts ("Content Mode") and tracking this in #35. I like your change, but cannot formally merge it because of the "undetectable" implications to the brand :)

Here's what we should do: you can configure your IDE with Prettier so systems always minimize changes. I will not merge the pull request here (too many code changes, and there's a reason for the current UI to co-exist), but your fork of the repository has your UI and University-focused changes.
You should maintain and publish your version whi (I use Vercel, free plan, and you can give it a domain), which focuses on students and UI, and I will continue modularizing and cleaning the base code, and adding core features. If you have core features and improvements, I'll merge them into the master code base, otherwise, you can maintain them. Sounds good?

willscottrod:
Ah, makes sense.

Yeah I was looking at it like "I didn't change that much". Turns out it is the formatter. I've used Prettier before but currently using Eslint so that's where that issue lies. Will change the formatter on my end for this project to keep consistency.

On the note of the toggle for the bigger text box, that sounds like a small task so I'll focus on adding that to my version so there's options in the UI.

My 2 cents on the Community Templates. The ideal goal is to have a marketplace style area that has different system messages that can be easily integrated for ultimate customization, however that could take a more intense hosting solution with the added database. I might start a small side project marketplace that prompts can be added to by users and voted up and down, and then I'll submit a pull request with an integration in the future that would allow people to select a prompt or upload a prompt.

So yup, I'll continue on my own fork for now and will definitely be in touch soon as your and my version progress.

rossman22590:
i am super interested in this will

enricoros:
Thanks Will. Before I park this pull request: are you thinking of maintaining and hosting (e.g., Vercel) your fork?

If so, I can add a link to your fork from the main readme, to send University users your way :) Lmk.

Sounds great for the prompts, your feedback is great for #35 .

willscottrod:
Hey

Yup, I'll self host this fork with a uni mode initially.

Gotta make some other modifications over this weekend then I'll flick you a link.

PR #42, opened 2023-03-31
Use Textarea to gather Custom system message instead of alert modal.

jondwillis:
From a UX perspective, `alert` inputs can be all over the place, but in my case, not good! This PR inlines the Custom systemMessage gathering.

- [x] Use Textarea to gather Custom system message instead of alert modal.
- [ ] Confirm that switching after already having started a chat works correctly for new inline "Custom" behavior.

Preview screenshot:
image

### DIFF
--- a/components/ChatMessageList.tsx
+++ b/components/ChatMessageList.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import { shallow } from 'zustand/shallow';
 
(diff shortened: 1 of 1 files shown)


enricoros:
Great PR, thanks for contributing this. Much better UX!!!

Once set, the custom prompt can be edited in the conversation ("show system text" in the menu), but edits won't be persisted to disk or reused if you clear the chat. Looking for feedback (#35 ) on how to make any preset editable as well.

PR #45, opened 2023-04-02
Made the default height of the input box 2 instead of 5

nilshulth:
I made this since the input box took up so much space on a small mobile screen, hard to see the output


enricoros:
Hi, made this comment.
image

enricoros:
Note: this commit went in because part of the other pull request. I've set the rows from 5 to 4, as 2 had bad spacing on desktop. 4 looks great on mobile too, where there are 4 buttons on the left side.
Thanks!

PR #51, opened 2023-04-03
Less message padding in markdown

nilshulth:
Removed unwanted padding at the top and bottom of the chat messages formatted with markdown. Had to override the CSS formatting of the p element. Also did adjustments to the width of p and table elements so that they display more nicely if they are very wide.


enricoros:
Thanks, applied! I've just asked GPT4 for what it would do to have a styling consistent to GitHub's Markdown rendering (I believe it to be good, but I'm not sure - that's my assumption).

@nilshulth: here's what it had to say. (I used GPT4 with drag&drop files to let it edit)
https://paste.gg/p/anonymous/b63477ac81214093aef6bfab3cc32c44    <-- scroll to bottom

Issue #59, opened 2023-04-06
In-line Free Scroll toggling 


jondwillis:
A common UX affordance is to enable free scroll and add a “⬇️ scroll to bottom” floating action button that disables free scroll. optionally, this new behavior could be disabled with another setting/env var.

enricoros:
Just tested this. A bit tough at the edges, but should be the default. Will need to do some cleanup, but thanks for the patch!

enricoros:
I admit the FAB is not there yet - and the scrolling is just okay, sometimes on mobile is not optimal. The browser-standard way of doing it, is to use a flexbox with row-reverse as the display mode. So that the message[0] is at the bottom and browsers go to it.

However one day we may re-migrate to the 'scrollToBottom()' way of doing it every time there's a new token..

PR #60, opened 2023-04-06
Automatic free scroll toggling and scroll to bottom floating button.

jondwillis:
https://github.com/enricoros/nextjs-chatgpt-app/issues/59

### DIFF
--- a/components/ChatMessageList.tsx
+++ b/components/ChatMessageList.tsx
@@ -1,62 +1,89 @@
 import * as React from 'react';
 import { shallow } from 'zustand/shallow';
-
-import { Box, List } from '@mui/joy';
+import { Box, List, Button, Typography, Switch } from '@mui/joy';
 import { SxProps } from '@mui/joy/styles/types';
-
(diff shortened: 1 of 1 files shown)


enricoros:
**I love this**! Need to do some cleanup for mergind this (somehow the PurposeSelector is gone, and the buttons appear to the far right of the screen (very right, see below).

But this is brilliant!
image

enricoros:
I've implemented a pure CSS solution to this (turns out that with flex-direction: 'column-reverse', it defaults to sticking to the bottom, unless the user has moved the scrollbar. The browser does the implementation basically.

However lost the smooth scrolling that was there before (each new line is immediately displayed). Removed the "Free Scroll" option too. Big win.

PaRaD1SE98:
In mobile Safari, it is not going down properly for some reason.

https://user-images.githubusercontent.com/78383353/235801293-74d456c5-8a51-4f38-92a5-b79b050562ed.mp4

PR #66, opened 2023-04-07
filter input for purposes

harlanlewis:
Adds a filter text input above system purposes, matches purpose.title and purpose.description. Mostly useful for folks who have expanded the SystemPurposes list.

I think this is useful as-is, but room for improvement:
- bug: the selected purpose should always be within filtered set. in this PR, it selects purpose[0] (ie, the first rendered item in the grid), but requires a couple keydown events before it works correctly.
- enhancement: fuzzy search or LLM process the input to do purpose matching (rather than literal string match), especially if systemMessage is included in search space
- bigger idea: Skip the up-front purpose switcher and use LLM for purpose matching based on the user's initial chat message (if user submits a sciency input message, first determine message type (Scientist), then send Scientist system message prompt, then submit user's input as first chat message)

localhost_3000_
localhost_3000_ (1)
localhost_3000_(iPhone SE)
localhost_3000_(iPhone SE) (1)

### DIFF
--- a/components/util/PurposeSelector.tsx
+++ b/components/util/PurposeSelector.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 
-import { Box, Button, Grid, Stack, Textarea, Typography, useTheme } from '@mui/joy';
+import { Box, Button, Grid, Input, Stack, Textarea, Typography, useTheme } from '@mui/joy';
+import { Clear } from '@mui/icons-material';
 
 import { useActiveConfiguration } from '@/lib/store-chats';
 import { SystemPurposeId, SystemPurposes } from '@/lib/data';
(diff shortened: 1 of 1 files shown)


enricoros:
Began reviewing for merging of the branch.

Few notes as I merge:
1. There is 1 single line of code that brings the build time of the app from 1 second to 9 seconds :)
image
This is the culprit (it bit me in the past): ```import { Clear } from '@mui/icons-material';``` --> replacing with the ```import ClearIcon from '@mui/icons-material/Clear';``` version.
What happens is that somehow the packager imports all of the material icons in that shorter version. (I love that shorter version by the way). This is after replacing the line:
image

I'm performing a couple more changes, little touches, for style. There's a bug I'm chasing which is a nitpick (the tile will be activated at the "next keystroke"), but to fix it I'm modifying a lot and opening more issues :)

Track progress here https://github.com/enricoros/nextjs-chatgpt-app/commit/fcfc0be269f8452195c8e5258b0e56ddc9284659

enricoros:
Done - please test it. *I've placed it under a setting in the Settings menu* <- please Activate the setting
Fixed the bug that you reported and all I could find.
image
image

Very well done, I love the usability you added here. Big win for the app. Love it, and great job in matching the style.

harlanlewis:
Tested it a little, looks great! I'll be using it quite a bit going forward. Love the improvements you made and impressed by how quickly you pulled it in. Thanks!

PR #69, opened 2023-04-10
added manifest.json for PWA / saved to home screen

harlanlewis:
adds manifest.json, a step toward PWA / opens as an app rather than a tab.


enricoros:
Thanks @harlanlewis, I was considering to add the full next-pwa structure, but don't have a good brand and icon yet. I guess we should solve that too. This is the vercel icon right?
Question: no web workers added - is this the intent?

harlanlewis:
Exactly, Vercel icon - just higher resolution versions of the favicon to avoid any brand suggestions/decisions. The goal of this PR was just to make it feel visually more at home when saved as an app, specifically:
- theme that matches app toolbar (in light mode)
- ios status bar style
- an app icon (any icon)

Also yes, omitted service worker. I don't have a vision for how offline might work with OpenAI API, or what you might have in mind for features like conversation history and custom purposes. So limited scope to manifest for now.

Absolutely no offense taken if this PR isn't baked enough yet :)

enricoros:
I have a concept for the Icon that I'd like to try. Let's keep this PR open, and as soon as I have a better icon, I will merge icon and your manifest. Thanks for providing it.

The advantage of the full next-pwa treatment is that it would be fast to load, display full screen on mobile, and enable "share to" from your phone. Would improve UX. At the same time I don't want to add a heavy service worker (it is) which also makes local development quite harder (there's something about registration/deregistration of web workers that tends to interfere with reloads/new code).

enricoros:
Merged. Better to have this in, than to not have it. Sets an icon and a name, but don't have better ones ready to land.
Thanks @harlanlewis

enricoros:
Looks really good @harlanlewis 

image

Once installed, it's really good-looking!

PR #74, opened 2023-04-11
add missing config to api publish

dario-piotrowicz:


enricoros:
Thanks for the lead @dario-piotrowicz , I've merged and adapted that function to work on edge runtime.

PR #75, opened 2023-04-11
Add setting to change speech-to-text input language.

tboydston:
A couple points regarding this pull request. 

1. The language list is not exhaustive and all languages may not be supported by all devices. I have tested English, German, Spanish, and Chinese on Chrome. I believe it should fall back to English on unsupported languages. 
2. The handleTextToSpeechLangChange function in SettingsModal runs a window.location.reload() to refresh the language input which seems a little dirty. Perhaps you can suggest an alternative.


enricoros:
Hi Tracey, love the patch! Somehow tho most of the patch is doing a style change, changing spacing and newlines and that stuff. Not a problem per se, but it will then be reverted back when I edit the file.
For style I'm using the "prettier" package, and the configuration fila is on the root of the project.
Any chance you can resubmit the pull request only with changes that relate to the feature?
To the feature itself: I love it, it's very needed.

enricoros:
I worked a long time on this today, but from the patch, it's hard to understand where the changes actually are.
There are more than 600 lines changed, of which at least 50% are just arbitrary modifications to the style of the project.
Due to quality I cannot merge. Please resubmit a minimal change, or I'll have to reimplement it from scratch (faster than reviewing 600 lines).

tboydston:
Sorry about that. I use prettier as well and I am not sure why my IDE reorganized so many things. I will clean it up and resubmit.

enricoros:
I'm not sure as well, I saw others submit the same styles changes. Can you try to load the /.prettier file? Maybe that sets it up.
I appreciate this work, in the near future, the same configuration options can change the language of the app, or voice output (coming soon).

tboydston:
There is no /.prettier file in the repo only a /.prettierrc file which it picks up because I can see the IDE enforcing the "trailingComma": "all", setting. 

I cloned a fresh version and ran "npx prettier components/dialogs/SettingsModal.tsx" and the formatting changes suggested match the ones I submitted in my pull request. This suggest to me that maybe there are some missing formatting settings in order to get everything to agree. 

Things like this the way you have the getters and setters organized in this section should require a special setting or exception to prevent prettier from trying to reorganize them. 

image

prettier really wants them to be. 

image

Feel free to let me know if I am missing something or have something configured wrong.

tboydston:
> I'm not sure as well, I saw others submit the same styles changes. Can you try to load the /.prettier file? Maybe that sets it up. I appreciate this work, in the near future, the same configuration options can change the language of the app, or voice output (coming soon).

I was looking into the audio output as well. The use case I am exploring is language learning by holding a simulated conversion in a second language. This is probably not where this should be discussed but it may be worth exploring integrating an external text-to-speech and speech-to-text provider at least as an option. The device output is kinda roboty but a lot cloud based text synthesis is really life like.

enricoros:
> Feel free to let me know if I am missing something or have something configured wrong.

Thanks, I apparently don't know how this works well :) I use WebStorm as the IDE, and it picked up the settings of Prettier, but did not execute prettier. If I do it, indeed, it formats all the code. So you're right and I'm wrong.

How can we get to the patch without having to rewrite the whole file? I need to review the differences pretty much - maybe disabling Prettier completely is the solution? (or telling prettier to not disrupt existing whitespace)

Quick feedback after a short review:

 - your select for the language could apply to the whole lang-dependent app features: **mic input**, **audio output**, and **translations** -- you should consider calling it "app language" or "Preferred language", so that we'll use that as the source of truth

 - the options could be separated in a smaller structure, (Object structure, like { en_US: 'English (US)', .... }, or List structure such as [{code:'en_US', name:'English (US)'}, ...], and injected in the UI with a {languages.map(lang => <Option key={lang.code} value={lang.code}>{lang.name}</Option>  <-- this way the UI file stays slimmer

Thanks again

tboydston:
> > Feel free to let me know if I am missing something or have something configured wrong.
> 
> Thanks, I apparently don't know how this works well :) I use WebStorm as the IDE, and it picked up the settings of Prettier, but did not execute prettier. If I do it, indeed, it formats all the code. So you're right and I'm wrong.
> 
> How can we get to the patch without having to rewrite the whole file? I need to review the differences pretty much - maybe disabling Prettier completely is the solution? (or telling prettier to not disrupt existing whitespace)
> 
> Quick feedback after a short review:
> 
> * your select for the language could apply to the whole lang-dependent app features: **mic input**, **audio output**, and **translations** -- you should consider calling it "app language" or "Preferred language", so that we'll use that as the source of truth
> * the options could be separated in a smaller structure, (Object structure, like { en_US: 'English (US)', .... }, or List structure such as [{code:'en_US', name:'English (US)'}, ...], and injected in the UI with a {languages.map(lang => {lang.name}  <-- this way the UI file stays slimmer
> 
> Thanks again

My suggestion would be to pick a style guide and run prettify on the whole project so that it is easier for others to contribute. If there are any specific blocks of code you would like to keep you can use the /* prettier-ignore */ comment to save their formatting. Also, if you are going to use es lint and prettier you need to install the prettier module for es lint and add it to the "extends" in the config otherwise there may be conflicts( see: [here](https://nextjs.org/docs/basic-features/eslint#usage-with-other-tools)  ). 

Let me know if that is the path you want to go down. Otherwise I will try to make the changes by disabling prettier.

enricoros:
> Let me know if that is the path you want to go down. Otherwise I will try to make the changes by disabling prettier.

Please go ahead and make the fixes disabling - I'd rather have them in sooner. I have a few large refactorings pending, and want to reset all files with prettier soon.. but not yet. You make a good point about contributions.

If you can make the patch disabling, I'll merge that in first, and then make the massive commit.

enricoros:
Hi @tboydston let me know if you have any updates on this, otherwise I'll go ahead and recreate something similar to yours.

tboydston:
I will try to make these changes today.

tboydston:
Ok, updated. Main changes. 

1. Updated the code formatting to match yours. 
2. Moved the languages to a separate file so they can be imported as needed. 
3. Improved the language list so language names are in their own language instead of English. 
4. Changed the setting variable name to preferredLanguage so that it can be used in other areas of the application. I left the setting label as "Speech Input Language" in the form because that is all the language changes at the moment. 

I don't know why the Vercel deployment has failed and whenever I click on the "Details" button Vercel just gives me a 404 page.

enricoros:
Thanks @tboydston I'll work on mainlining this tonight, expect to have it integrated by tomorrow.

enricoros:
(no worries about conflicts, that's on me, easy peasy)

enricoros:
@tboydston - merged! Great work, thanks. I've promoted it to the top setting, as it's very important and can be the base of app translation and selection of speech-language (not in Elevenlabs, yet, but future). Using the default language from the browser, so even without changing it in the settings, the mic should adapt (where supported).

PR #80, opened 2023-04-12
toggle api key

harlanlewis:
2 changes:

- toggle visibility of API key value using with `Input endDecorator={eyeball icon button}`. I shuffle a few keys associated with different accounts so it's useful to spot check this. The decorator is only shown while the input has a value.
  - I have not tested how this works with API key as env variable - just local storage
- change label of Settings `Close` button to `Save`, as its effective function is similar to a submit button

localhost_3000_


enricoros:
Merging this in a second. The 'save' will be reverted, based on the fact that the dialog has a 'x' button.
With 'Save' one would have the mistaken idea that by pressing the 'x' their changes (e.g. changing the API key) would be undone. But save is actually just a close (does not store the new values) - so I'll keep it as close for now.
The rest is great!

enricoros:
Thanks @harlanlewis, looks great, and the implementation is very clean.

PR #81, opened 2023-04-13
Enabled stream option for Text decoding

shinkawk:
A quick bug fix.

https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder/decode

Added `stream `option to TextDecoder.decode in order to prevent malformed multi-byte characters like on in below. (Happens more often when deployed to vercel than running it on local)
 
スクリーンショット 2023-04-13 9 44 58 (3)

### DIFF
--- a/lib/ai.ts
+++ b/lib/ai.ts
@@ -51,7 +51,7 @@ export async function streamAssistantMessage(
 
         if (done) break;
 
-        incrementalText += decoder.decode(value);
+        incrementalText += decoder.decode(value, {stream: true});
 
         // there may be a JSON object at the beginning of the message, which contains the model name (streaming workaround)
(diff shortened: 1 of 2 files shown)


enricoros:
Thanks, @shinkawk 🙏  - very good patch!
How did you find out?! I had no idea about this phenomenon.

shinkawk:
Hi @enricoros,

No problem at all!

Found this issue when I was throwing some questions in Japanese. 
Most of the time, the streaming response body was returning in small bits and there was no problem decoding them. But for some reason, the word "Arigatogozaimasu" (Thank you in Japanese) was always split into 2 big chunks and the middle word "za" was cut in half (thus causing it to be malformed).

PR #87, opened 2023-04-16
run code in Replit / Codepen

harlanlewis:
**Edit 2 - WIP:** Replit is usable as-is, but isn't actually passing the title and code (need to be manually pasted)

**Edit: this PR was initially just for [Codepen](https://codepen.io/) (HTML/CSS/JS), but expanded it to [Replit](https://replit.com) for C#, Java, and Python**

--

Run generated code in Codepen (frontend) or Replit (backend)

- Adds an "Open in Codepen" / "Open in Replit" - opens in new tab, no account required
- Improves code block button readability when overlaying text with `backdrop-filter: blur`
- Use Prism for language detection fallback to the "startsWith" conditionals.

Initial state (just for context, no button shown)
Screen Shot 2023-04-15 at 9 07 23 PM

Hover code block:
Screen Shot 2023-04-15 at 9 07 44 PM

Hover CodePen button:
Screen Shot 2023-04-15 at 9 07 56 PM


harlanlewis:
A couple examples:

**Replit Python:**
screenshot-2023-04-15 22-04-55
screenshot-2023-04-15 22-05-31

---
Codepen:
screenshot-2023-04-15 22-13-21
screenshot-2023-04-15 22-13-39

enricoros:
@harlanlewis: this is absolutely great. And the patch is relatively small!
Great job, I'm maintaining a branch `variant-code-execution` that uses Sandpack to run some web code, but your approach seems even more effective.

Codepen: do you know if it's <iframe>'able or embeddable in other ways? Showing it inline could be great (but even the button makes so much sense)

Replit: works, and seems embeddable. - see https://docs.replit.com/hosting/embedding-repls

@harlanlewis: reviewed, and your code is of good quality. Can I proceed with the merge, or how else can I be of your help? 
 (need any UX cleanup, such as: "Code execution: [ ] off, [ ] inline, [ ] external" on the Settings menu or similar?)

harlanlewis:
Please feel free to merge if you think it’s ready. Improvements can come iteratively. 

I don’t think a settings option is necessary for this - there is no 3rd party tracking introduced, no accounts required for the services, and the UI footprint in the chat response is minimal. If you can get embeds working, internal vs external might be a useful toggle. Currently Codepen button is also shown for the Designer purpose SVGs, so you can see how it feels having both internal vs external render buttons side by side. 

Codepen actually has a couple of different embed methods:
- [oEmbed](https://blog.codepen.io/documentation/oembed/) 
- [Embed Builder](https://blog.codepen.io/documentation/embedded-pens/) 
- iframe

My original goal was to embed rather than link out, but wasn’t able to make it work without multiple manual steps.

harlanlewis:
At a glance, there are some cleanup issues - hasSVG shouldn’t be part of the check for displaying the Replit button. You are welcome to make any improvements you see, I will probably not be able to touch this until tomorrow.

enricoros:
Very well done. I'll take a look if there's anything to polish, but looks great.

PR #93, opened 2023-04-25
Rudimentary implementation of ReAct with Wikipedia and Calculation as first tools.

fredliubojin:
* Only triggered when in ReAct mode.
* Only display ReAct results in Console, not updating the conversation yet.


enricoros:
Incredibly efficient ReAct implementation. Thanks @fredliubojin 

image

PR #116, opened 2023-05-21
Prevent menus from hiding on small window height

typpo:
There's a bug where if the window height is less than the menu height, the menu disappears.
big-agi-dropdown-bug

Removing the `placement` prop and setting maxHeight seems to fix the issue.  It seems like Joy UI is already smart enough to place the menu correctly.
big-agi-dropdown-fix


enricoros:
Thanks! Great fix for a big problem!

enricoros:
@typpo Thanks Ian - I've also propagated to the `next` branch which is almost ready to be merged into main, and will bring some flexibility to the application structure (including tRPC and more).

PR #121, opened 2023-05-29
Update models.ts

rossman22590:
updated models


rossman22590:
updated to new Prodia models :)

PR #127, opened 2023-06-16
feat: add moderation 

### LINKED ISSUE #55: Optional pre-flight check Moderation API to avoid abuse
[According to OpenAI's usage policy](https://openai.com/policies/usage-policies), it may be possible to have action taken against an API account if enough violations occur. 

>If we discover that your product or usage doesn’t follow these policies, we may ask you to make necessary changes. Repeated or serious violations may result in further action, including suspending or terminating your account.

To help avoid that, add an option to [enable pre-flight moderation API queries](https://platform.openai.com/docs/guides/moderation/quickstart).

danbalarin:
- [x] Add new API endpoint `/api/openai/moderation`
- [x] Add according settings value, so user can toggle this off
- [x] Add the required check on every chat message

fixes #55

### DIFF
--- a/pages/api/openai/moderation.ts
+++ b/pages/api/openai/moderation.ts
@@ -1,0 +1,50 @@
+import { NextRequest, NextResponse } from "next/server";
+import { ChatModerationSchema, chatModerationSchema, openAIAccess } from "~/modules/llms/openai/openai.router";
+import { OpenAI } from "~/modules/llms/openai/openai.types";
+
+async function rethrowOpenAIError(response: Response) {
+    if (!response.ok) {
+      let errorPayload: object | null = null;
+      try {
+        errorPayload = await response.json();
+      } catch (e) {
+        // ignore
+      }
+      throw new Error(`${response.status} · ${response.statusText}${errorPayload ? ' · ' + JSON.stringify(errorPayload) : ''}`);
+    }
+}  
+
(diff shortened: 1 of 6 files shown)


enricoros:
Thanks for working on this! Will process it later today. The quality of the code seems good, I may move the moderation to tRPC if possible.

When the feature is off, will it impact at all the current workflow for existing users (latency, data, ..)?

danbalarin:
Cool, I have little to no experience with tRPC, so I'd be glad if you could move it. 

When the feature is off it doesn't affect the user as much as one additional condition to be processed by CPU 😄. And when it's on, there is one request before each new message, but according to the docs, these requests are not calculated towards some payment or anything. But that means, you have two requests in a series, potentially increasing latency and data usage. On the other hand, when you have a conversation with tens of replies, the additional data usage in minor.

danbalarin:
Also, I picked this issue at random, seemed like good first issue and I had some time to kill, so there is no need to rush anything from my POV 😄

enricoros:
@danbalarin Merged!

Good placement of the moderation flags and checking for the user input.

I'm defaulting `moderationCheck` to `false`, so there's no additional roundtrip for current users - however, I see the benefit for sysadmins to turn it on by default in their managed installations. maybe the default should be centralized to a config file, or tied to env variables, along with other defaults.

Thanks for your contrib!

PR #128, opened 2023-06-19
Loading System Purposes from JSON file, overriding default data.ts va…

syberkitten:
Loading System Purposes from JSON file, overriding default data.ts values,
    allows to extend without changing sourcecode.
- Add Teacher and Default system purposes, update API host, and load systemPurposes from JSON file
- Add .env.* to .gitignore
- Modify src/data.ts to load systemPurposes from systemPurposes.json
- Add Teacher and Default system purposes in systemPurposes.json


syberkitten:
Hi Guys,
Amazing project!!!

by the way It may be preferred to put the systemPurposes.json inside the env vars section
such as:
OVERRIDE_SYSTEM_PURPOSES="systemPurposes.json"
so that it will be overloaded (extended) only when it exists.

Basically this idea comes from the fact that I'm using BigAGI in tandem with another
project and i want to get the latest features while still extending it
without having to have code collisions.

Ideas are welcome and I'll be glad to assist... )

enricoros:
Thanks @syberkitten, good idea, smart implementation. Do you know whether the loading code runs at build time or runtime, and client or server? that 'require' is dark magic to me :)

enricoros:
Another question @syberkitten: what do you think if we had the current SystemPurposes inside the JSON, with maybe 1-2 as fallback in case the JSON is loaded incorrectly? 

This would allow for greater customization, since the current impl wouldn't allow removal, only add/override.

Also, there's an 'invisible' list of IDs in the store-purposes... maybe that should be loaded from a JSON the first time..

syberkitten:
Thanks for the feedback @enricoros 
As to the first question the require('../systemPurposes.json') dark magic is used on the server side during the build process. 

When the code is bundled and built, the server will attempt to load the `systemPurposes.json` file and merge its contents with the `SystemPurposes` object which should be available on both the server and client sides after the build process. 

As to the second question I'll give it a further dive and get back to you :)

syberkitten:
I would also love to do that same for the branding if that makes sense.

There is also a bug where if the OpenAI api key is provided as an ENV variable, none of the models are loaded,
I've created a default set of models to solve it locally, but it is just a workaround, and i will address this issue seperately.

enricoros:
> @enricoros can you please specify which fallback purposes would you have as fallback??

Good point, no need for those. Sorry for the lag, I had a few other changes to bring in before this. I've tried and it works.

DMTarmey (approved):

DeFiFoFum:
I'm also a fan of being able to drop in agent purposes with a JSON and to be able to back it up easily and share. 🔥

syberkitten:
Ok, I've made final changes to the Pull Request including these features:

1. Moved the system-purposes.json (now it's called such to be more in par with naming conventions)
the file is now inside a new folder called '**customization**' so not to clutter the root. - other customizations can go there later on such as for branding etc....
2. @enricoros system-purposes.json now has Settings key, with the hidden purposes IDS which are loaded from the json if they exist otherwise a default value inside store-purposes.ts will be used
3. As a game I've added another purpose, you heard of DAN? it's now hidden and can be enabled with the normal "Edit" when selecting a purpose.
4. I've Enlarged a bit the purpose boxes so that a longer title will fit when editing them, perhaps it needs some more adjustment to look better, I'm not strong on the react layout side, yet....

Let me know if you have other remarks or ideas, looking forward to see this in as I've already started working on more stuff, so next PR is coming soon.... :)

syberkitten:
Hi @enricoros and @DeFiFoFum 
I've added another fix to make sure that default "hiddenPurposeIDs" work even if the system-purposes.json is missing.

syberkitten:
Hi @DeFiFoFum 
Yes for new users and old, we need to document this feature, so it'll be clear how to change such Personas.
so I will add a section to the readme.md or equivalent section.

Other then this, I wish is to be able to customize both branding as well.
a kind of a "snap in" plugin system that will allow the project be used for different use cases, would be amazing.

Some of the features here (for my use case) are not really needed, but they are "backed"
and make the project more complex to maintain.... I was actually pursuing just a simple chatGPT
client for React JS when I found out about BigAgi.

enricoros:
@syberkitten I started merging this, but then went down a rabbit hole of trying to make an editing system. So can't merge yet as I wanted.
I'll probably have to restart on this from scratch as I accumulated too much experimental code but your request is quite straightforward.

(I started using Zod for type validation, then did an overlay (instead of append) of the purposes, then renamed to personas, then changed the data structure, etc etc.. -- basically, tried to boil the ocean and haven't done yet)

enricoros:
Hi @syberkitten, closing this as will be tracked with a new system of making personas dynamic. #35

PR #140, opened 2023-07-08
feature: A command to send messages as a system

JustMrPhoenix:
I have implemented a new command to allow users to send messages labeled as "system". This is intended to allow users to test out behavior before using it directly with API. The system messages allow for better "steerability".


enricoros:
Thanks @JustMrPhoenix - also added /a for adding Assistant messages.
In your original vision for the feature, did you just want to add the message (assistant / system) or add them and get the response?

JustMrPhoenix:
For system messages it makes sense to generate a response after it was sent since system implies a new message passed to the assistant. As for the assistant messages I guess it doesn't make sense to request a response for them. The way I used to approach that in the past is just edit assistant's message with the text I want

PR #155, opened 2023-08-27
Enable User ID in environment and advanced model configuration

peterbsmyth:
the general concept is that if a `USER_ID` is set then it will be applied to both OpenAI and Helicone


enricoros:
Thanks for submitting. I'll be ooo for the week (yearly travels), will merge after.

enricoros:
See #180

PR #157, opened 2023-08-29
Add chat drawer indicator for num conversations

DeFiFoFum:
Simple indication of the number of chats currently being used out of the max chats.

The next goal is to add an edit button for the `ChatDrawerItems` be able to multi-select the chats to remove.

Screenshot 2023-08-29 at 5 31 39 PM

### DIFF
--- a/src/apps/chat/components/applayout/ChatDrawerItems.tsx
+++ b/src/apps/chat/components/applayout/ChatDrawerItems.tsx
@@ -42,9 +42,10 @@ export function ChatDrawerItems(props: {
   }), shallow);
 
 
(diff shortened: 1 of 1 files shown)


vaibhavard:
This would be very helpful

Ashesh3:
This PR removes the max limit for Chats: https://github.com/enricoros/big-agi/pull/158

If the above gets merged, we can safely close this PR :)

enricoros:
@DeFiFoFum hey, I've merged this.
As we also merged the infinite chats support (IndexedDB storage), I've only kept the counter, but removed the MAX.

Thanks!

Issue #137, opened 2023-07-04
Enhanced Chat History Management 


syberkitten:
This is a feature request not an issue.

It would be great to be able to save all the chats history and reload it,
this way will allow much more flexibility to use the system for research and work,
with having different "Contexts" of chats.

Further I would say that each chats history should first not be limited to 20 items (no limit is better)
And also the ability to group chats history.

For example when I'm working on code, I might open 5-10 chats which are related to one topic of work,
then I start another project, and can open another 20 chats.

Currently I'm very limited in this scope, so the export import is just a workaround / first steps,
the main idea is that you can manage different Dialogue groups and of course export / import them as well.

Last but not least, search within all these History Chat history Dialogue History, 
I don't understand why nobody did this yet, it's a must - go try find a sentence in 20 chats manually....

syberkitten:
This is how ChatGPT4 Summarizes this request:

Enhanced Chat History Management

1. **Chat History Saving and Reloading**: The ability to save and reload chat histories would significantly improve the system's flexibility, particularly for research and work purposes. This feature would allow users to maintain different "contexts" of chats.

2. **Unlimited Chat History**: Currently, chat history is limited to 20 items. Removing this limit would provide a more comprehensive view of past interactions.

3. **Chat History Grouping**: The ability to group related chats would be beneficial. For instance, when working on a project, a user might open multiple chats related to a single topic. Grouping these chats would help keep related discussions organized and easily accessible.

4. **Chat History Management**: The ability to manage different dialogue groups, including exporting and importing them, would provide greater control over chat interactions.

5. **Chat History Search**: A search function within the chat history would greatly improve usability. This feature is essential for quickly locating specific information within numerous chat interactions.

These enhancements aim to provide a more user-friendly and efficient chat experience, particularly for those using the system for complex projects or research.

tmikaeld:
I'm in the same boat, I use it mainly for coding.

It would be great if:

- We could have as many chats as we need.
- Could have multiple tabs with chats (That can be named).
- The ability to purge a whole tab (With confirmation)

enricoros:
> I'm in the same boat, I use it mainly for coding.

Same - and for life advice, workout plan, to go IRL to restaurants and pick wines, and now also to chat with public personas ;)

> We could have as many chats as we need.

[...]

PR #158, opened 2023-08-30
Switch from LocalStorage to IndexedDB for Chat History Storage

Ashesh3:
Partially Resolves: #137

This PR addresses the limitation of storing only 20 conversations in the chat history due to the max storage limit of LocalStorage. The solution proposed is to switch to IndexedDB for storing chat conversations, which has a higher storage limit and thus will allow for an unlimited number of conversations to be stored.

Key changes include:

1. Addition of `idb-keyval` package to interact with IndexedDB.

2. Removal of `MAX_CONVERSATIONS` constant and related logic that limited the number of stored conversations.

3. Modification of the `useChatStore` state to use IndexedDB for storage instead of LocalStorage.

4. Update of the `ChatDrawerItems.tsx` and `ChatMenuItems.tsx` components to remove the restrictions on maximum conversations.

By implementing these changes, users will no longer face the restriction of a limited number of stored conversations, improving the user experience and functionality of the chat feature.

Note: This PR includes a version bump to 3 in `store-chats.ts` to reflect the switch to IndexedDB.

image

Storage limit on IndexedDB:

- **Firefox:** 10% of the total disk size. (up to 10GB)
- **Chrome:** 60% of the total disk space. (No max limit)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24,6 +24,7 @@
         "@vercel/analytics": "^1.0.2",
         "browser-fs-access": "^0.34.1",
         "eventsource-parser": "^1.0.0",
+        "idb-keyval": "^6.2.1",
         "next": "^13.4.19",
         "pdfjs-dist": "3.9.179",
         "plantuml-encoder": "^1.4.0",
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "@vercel/analytics": "^1.0.2",
     "browser-fs-access": "^0.34.1",
     "eventsource-parser": "^1.0.0",
+    "idb-keyval": "^6.2.1",
     "next": "^13.4.19",
     "pdfjs-dist": "3.9.179",
     "plantuml-encoder": "^1.4.0",
--- a/src/apps/chat/components/applayout/ChatDrawerItems.tsx
+++ b/src/apps/chat/components/applayout/ChatDrawerItems.tsx
@@ -1,12 +1,12 @@
 import * as React from 'react';
 import { shallow } from 'zustand/shallow';
 
-import { Box, ListDivider, ListItemDecorator, MenuItem, Tooltip, Typography } from '@mui/joy';
+import { Box, ListDivider, ListItemDecorator, MenuItem, Typography } from '@mui/joy';
 import AddIcon from '@mui/icons-material/Add';
 import DeleteOutlineIcon from '@mui/icons-material/DeleteOutline';
 import FileUploadIcon from '@mui/icons-material/FileUpload';
 
-import { MAX_CONVERSATIONS, useChatStore } from '~/common/state/store-chats';
+import { useChatStore } from '~/common/state/store-chats';
 import { setLayoutDrawerAnchor } from '~/common/layout/store-applayout';
 import { useUIPreferencesStore } from '~/common/state/store-ui';
 
@@ -44,7 +44,6 @@ export function ChatDrawerItems(props: {
 
   const hasChats = conversationIDs.length > 0;
   const singleChat = conversationIDs.length === 1;
-  const maxReached = conversationIDs.length >= MAX_CONVERSATIONS;
 
   const closeDrawerMenu = () => setLayoutDrawerAnchor(null);
 
@@ -67,8 +66,6 @@ export function ChatDrawerItems(props: {
     if (!singleChat && conversationId)
       deleteConversation(conversationId);
   }, [deleteConversation, singleChat]);
-
-  const NewPrefix = maxReached && <Tooltip title={`Maximum limit: ${MAX_CONVERSATIONS} chats. Proceeding will remove the oldest chat.`}><Box sx={{ mr: 2 }}>⚠️</Box></Tooltip>;
 
   // grouping
   let sortedIds = conversationIDs;
@@ -98,9 +95,9 @@ export function ChatDrawerItems(props: {
     {/*  </Typography>*/}
     {/*</ListItem>*/}
 
-    <MenuItem disabled={maxReached || (!!topNewConversationId && topNewConversationId === props.conversationId)} onClick={handleNew}>
+    <MenuItem disabled={!!topNewConversationId && topNewConversationId === props.conversationId} onClick={handleNew}>
       <ListItemDecorator><AddIcon /></ListItemDecorator>
-      {NewPrefix}New
+      New
     </MenuItem>
 
     <ListDivider sx={{ mb: 0 }} />
--- a/src/apps/chat/components/applayout/ChatMenuItems.tsx
+++ b/src/apps/chat/components/applayout/ChatMenuItems.tsx
@@ -10,7 +10,7 @@ import FileDownloadIcon from '@mui/icons-material/FileDownload';
 import ForkRightIcon from '@mui/icons-material/ForkRight';
 import SettingsSuggestIcon from '@mui/icons-material/SettingsSuggest';
 
-import { MAX_CONVERSATIONS, useChatStore } from '~/common/state/store-chats';
+import { useChatStore } from '~/common/state/store-chats';
 import { setLayoutMenuAnchor } from '~/common/layout/store-applayout';
 import { useUIPreferencesStore } from '~/common/state/store-ui';
 
@@ -28,7 +28,6 @@ export function ChatMenuItems(props: {
   const { showSystemMessages, setShowSystemMessages } = useUIPreferencesStore(state => ({
     showSystemMessages: state.showSystemMessages, setShowSystemMessages: state.setShowSystemMessages,
   }), shallow);
-  const maxConversationsReached: boolean = useChatStore(state => state.conversations.length >= MAX_CONVERSATIONS);
 
   // derived state
   const disabled = !props.conversationId || props.isConversationEmpty;
@@ -82,13 +81,13 @@ export function ChatMenuItems(props: {
 
     <ListDivider inset='startContent' />
 
-    <MenuItem disabled={disabled || maxConversationsReached} onClick={handleConversationDuplicate}>
+    <MenuItem disabled={disabled} onClick={handleConversationDuplicate}>
       <ListItemDecorator>
         {/*<Badge size='sm' color='success'>*/}
         <ForkRightIcon color='success' />
         {/*</Badge>*/}
       </ListItemDecorator>
-      Duplicate{maxConversationsReached && ' (max reached)'}
+      Duplicate
     </MenuItem>
 
     <MenuItem disabled={disabled} onClick={handleConversationFlatten}>
--- a/src/common/state/store-chats.ts
+++ b/src/common/state/store-chats.ts
@@ -116,6 +116,23 @@ const storage: StateStorage = {
   },
 }
 
+function _migrateLocalStorageToIndexedDB(state: ChatStore) {
+  const key = "app-chats"
+  const value = localStorage.getItem(key);
+
+  // Check if migration has already been done
+  if (!value) return state;
+
+  // Mark migration as done
+  localStorage.removeItem(key);
+
+  // Migrate data to IndexedDB
+  const localStorageState = JSON.parse(value)?.state;
+
+  state.conversations = localStorageState?.conversations;
+  state.activeConversationId = localStorageState?.activeConversationId;
+}
+
 /// Conversations Store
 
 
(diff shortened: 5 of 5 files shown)


Ashesh3:
@enricoros Did you get a chance to review it? :) Hope you are enjoying your vacations!

enricoros:
Hi, this looks really good! I was on purpose ignoring requests to prioritize product brainstorming to code, but this looks like a great quality commit.

Do you think the chats can be preserved / migrated when switching? (As a one time operation)

Ashesh3:
> Hi, this looks really good! I was on purpose ignoring requests when coming back from vacation, to prioritize product brainstorming to code but this looks like a great quality commit.
> 
> Do you think the chats can be preserved / migrated when switching? (As a one time operation).
> 
> I'm asking because many people have valuable data in the chats and would be disappointed at the product if it wiped the chats at a certain point.

Hey, thanks for the comment, I would have used the [`migrate` option](https://docs.pmnd.rs/zustand/integrations/persisting-store-data#migrate) in the persist middleware if it were not for the fact that we are switching the entire storage backend itself. In this case, it will not be able to read the old data saved in local storage since the change itself suggests it to use Indexed DB going forward. 

What we can do here is, instead use a custom one-off migration to migrate local storage to indexed DB, it will check if there's existing data in local storage, if so, it will migrate it then clear it out from local storage so that the migration need not happen again. 

I've implemented the above-mentioned migration in this PR. Do review and let me know how you feel about this approach!

Note: Tested by manually setting local storage key `app-chats` from the current live version on this deploy preview and was successfully able to move that data to indexed DB.

Ashesh3:
@enricoros any updates on reviewing this? :P It'd be really nice to have 20+ chats, I constantly find myself having to delete chats 😅

enricoros:
I'm being very careful particularly on this, as I'm super scared to kill people's chats. Imagine the backlash! The code is definitely of high quality, and I want to merge it. Just being cautious for this, as it's potentially destructive.

### On the Chat Limit
 I am thinking of keeping the limit mechanism in place and raise it to 50. I have no idea what will happen for someone that goes to thousands of chats, and then the DB corrupts or the browser crashes or whatnot.
 
What if we keep the safety net? The product has no analytics, so we don't know how many people are at limit, but I'm pretty much at limit in all my installs :). However, raising 20 -> 50 could be a good safety net to keep in place.

There's a visual indicator on the chats list that can be activated when nearing the limit (turn it on with Preferences > Experiments > ON), and can help to visually see which chats are "mostly empty" and good to delete.

You may also consider the "MAX" limit as a limit that forces creativity and tidyness. My ChatGPT (I haven't used it since big-AGI) looked like an untify mess :) It's like having 160 chars for a tweet. 

So the question is: how do you feel to raise the limit to 50, as a "defensive/safety" measure, while switching to IndexDB?

### Side Effects / Other Questions
Support: I checked browser compatibility for IndexedDB and it seems good, but I haven't checked on a few browsers (Android/Chrome, iOS/Mobile Safari, MacOS/Safari). Do you think there's the chance some configuration will be unsupported?

Speed: haven't checked the speed difference of IndexedDB vs localStorage - is there any? 

Versioning: If this works well, we can migrate more things to IndexedDB, and use it properly as a structured DB - for instance AI personas definition, which really needs something like a DB. Do you think we need to "version" the name of the DBs/Tables?

enricoros:
Also, thinking of importing IndexedDB management functions (from the 'idb-keyval' lib, https://github.com/jakearchibald/idb-keyval) internally - so we can have error handling and more control. e.g. common/utils/idbUtils.ts. Wdyt?

Ashesh3:
> I'm being very careful particularly on this, as I'm super scared to kill people's chats. Imagine the backlash! The code is definitely of high quality, and I want to merge it. Just being cautious for this, as it's potentially destructive.

I totally understand, merging something which has the potential to destroy existing data should be thoroughly considered and examined. I have tested the code on 3 of my existing deployments and there were no issues migrating the chats and no data was lost. If someone uses the app for the first time, they use indexedDB in the first place, no issues for them too. I encourage you and others stumbling across this PR to copy their localstorage key to this deploy preview to see for themselves their data being migrated and not losing anything. 

> I am thinking of keeping the limit mechanism in place and raise it to 50. I have no idea what will happen for someone that goes to thousands of chats, and then the DB corrupts or the browser crashes or whatnot.

I would have to disagree with keeping the limit mechanism as the entire point of having chat history is that they serve as historical records to refer to in future, I take reference from every other provider, all of them instead of imposing limits on total conversations, instead aim towards improving chat management to make it easier for users to find any of their old chats. If someone hits the 50-chat limit, now every time they have to delete a previous conversation and spend extra time and effort to manually erase chats not needed anymore which is something I believe users should not be troubled with.  

Talking about DB corruption, it's as likely to happen as with local storage, both are ultimately managed by the browser itself and corruption to browser files will result in data loss in both cases, no exceptions. In terms of reliability, the way atomic transactions work in IndexedDB, in case a crash happens, the browser will preserve the state of the database just the second before crash. WhatsApp for example stores 100s of MBs of data in the IndexedDB and billions use it on a daily basis.


> You may also consider the "MAX" limit as a limit that forces creativity and tidyness. My ChatGPT (I haven't used it since big-AGI) looked like an untify mess :) It's like having 160 chars for a tweet.

I agree with you here, my ChatGPT history is probably cluttered with old conversations I no longer need but I still do not feel like deleting them. I feel it should be left on the user to decide, if they want to keep it organized or not. We should not force them into our own ways. if I want a quick chat, I do not sit and spend time finding a useless chat to delete.

> I checked browser compatibility for IndexedDB and it seems good, but I haven't checked on a few browsers (Android/Chrome, iOS/Mobile Safari, MacOS/Safari). Do you think there's the chance some configuration will be unsupported?

I am confident that there would be no issues combability wise, It's a fairly matured feature and used by major projects across the globe. It's supported in all browsers, on all platforms (Opera Mini (phone) does not support either localstorage or indexedDB). I am certain not a single user would be left out incompatible in scope of this project.

> Speed: haven't checked the speed difference of IndexedDB vs localStorage - is there any?

IndexedDB is slower by a few milliseconds, not a difference to be ever noticed by any user. 

> Versioning: If this works well, we can migrate more things to IndexedDB, and use it properly as a structured DB - for instance AI personas definition, which really needs something like a DB. Do you think we need to "version" the name of the DBs/Tables?

This is a good idea; AI personas would benefit from moving it out to IndexedDB. We already have versioning for the database, it will start at 3 once this PR is merged. We can also use a separate IndexedDB for AI personas and keep it separate and clean from the chat records.

Ashesh3:
> Also, thinking of importing IndexedDB management functions (from the 'idb-keyval' lib, https://github.com/jakearchibald/idb-keyval) internally - so we can have error handling and more control. e.g. common/utils/idbUtils.ts. This is so that we can change DB names, be in control of closing, apply some of the fixes in the upstream repo that haven't been merged. Wdyt?

No issues doing that but i ask myself if we really do need to reinvent the weel and copy their implementation when their library works best for our use case already. If any errors, although rare, do happen, the middleware will catch it and preserve the state accordingly is my understanding. I feel like we'd be unnecessarily adding to the codebase for something that isn't worth major improvement. Basically, keeping it simple and let the library do state management / error handling for us.

Also, currently the middleware allows us to detect and run [migrations](https://docs.pmnd.rs/zustand/integrations/persisting-store-data#migrate), in case we decide to add/remove/alter any field in the database. 

But yes, changing the DB name itself isn't simple. But again, how often are we going to change the entire name of the database where chats are stored?

enricoros:
@Ashesh3 great comments across the board. Gives me good confidence on the code. In the end I'm the only QA to ensure the production-ready quality of the App, but I feel that you get that and I'm very comfortable with the patch.

Apologies for not jumping on it already - I stayed away from code on purpose the last couple of weeks. I'll be hosting a friend for the next few days, but this will be the first commit you'll see.

I'll probably add back the 50 or 100 limit, but again, it's just a product choice, one of those "constraints that push creativity" kind of things. To be lifted quickly in time. But no worries about the code: I was actually looking at the merge last night and I'll add back the code where needed.

Thanks again, and look forward to this landing next week. If you're interested, I'm also working on the larger and long term planning for this project.

enricoros:
Found a bug that I'm trying to fix: if one opens the app (state is migrated*) and then closes it and opens again, all chats are lost.
The reason is that 'setItem' on idb is not called after the _migrate..() function restores the localStorage data.

Fixing this one before committing, otherwise there's the chance of data loss on an immediate refresh post migration.

Ashesh3:
> Found an issue: if the user opens the app (state is migrated) and then closes the window and opens again, _**chats are lost**_.
> 
> The reason is that 'setItem' on idb is not called after the _migrateLocalStorageToIndexedDB function restores the localStorage data.
> 
> Looking at the 'persist.ts' middleware implementation in Zustand, [it seems that there isn't any 'setItem()' after the onRehydrateStorage()](https://github.com/pmndrs/zustand/blob/deba6ed0d05b0dcb66c95177035ec8a061715745/src/middleware/persist.ts#L485C17-L485C17).
> 
> Also see [the comment a few lines above:](https://github.com/pmndrs/zustand/blob/deba6ed0d05b0dcb66c95177035ec8a061715745/src/middleware/persist.ts#L430C10-L430C10):
> 
> image
> 
> We seem to have the issue that migrate() is the place where to perform the load from localStorage, but migrate is not called (as you point out) because we swap the storage backend. The next place, `onRehydrateStorage` is not persisted to disk after updating the state, which may lead to data loss if the user reloads or closes the window before updating any state - since at this point the local storage values are gone.
> 
> @Ashesh3 Ideas?

How about this approach? We only clear localStorage after verifying that the IndexedDB contains the migrated data and that it matches the data stored in localStorage. So, before deleting localStorage, we confirm that the chat count in the state (after hydration) is the same as in localStorage. In this case, we can be confident that the migration has been successfully completed and persisted, and it's safe to remove the data from localStorage.

If a situation arises where the user does not interact with the app (and subsequently closes it), the state update may not be written to IndexedDB as expected. However, the next time the user uses the app and interacts with it, their chats will still be available in localStorage, and migration will be attempted again.

After the migration, we'll check if the chat object matches the one in IndexedDB. We could do this in the onFinishHydration function. Only then will we clear the localStorage key.

Otherwise, we could manually call the `setItem` function with the key and value to perform the migration.

What do you recommend?

enricoros:
Thanks for the two ideas. I tried looking into the `setItem`, but could not get visibility into that method. Possibly the first idea is best, but need to look at it a bit more.
By the way, the merge in progress is in the https://github.com/enricoros/big-agi/tree/Ashesh3-IndexedDB-storage branch, in the repo, I've not done anything major - just need to clear this issue first.

enricoros:
@Ashesh3 Thanks for your contribution here. This has been merged and tested thoroughly. 

The localStorage key is backed up to `app-chats-v2` once imported, and the updated code is using a trick to trigger a `migrate`-ion.

Very good help here, the app is better now.

tmikaeld:
This is awesome! Great commit and I love that you're both so careful ❤️ 

I still think it could benefit from categories/folders/tabs or similar, because now it's just one list short of chaos.
Plus, it would prevent huge DOM objects causing the browser to lag.

So with the 50/100-object limit proposed, maybe if we apply that to the categories, then anyone can just create a new category when it's been filled up.

And you're right about indexedDB not being limited, I have a database with millions of items and handling that without issues. It's only an issue if you apply string-based search or filtering on it, fetching keys is no issue at all no matter the size.

Ashesh3:
> This is awesome! Great commit and I love that you're both so careful ❤️
> 
> I still think it could benefit from categories/folders/tabs or similar, because now it's just one list short of chaos. Plus, it would prevent huge DOM objects causing the browser to lag.
> 
> So with the 50/100-object limit proposed, maybe if we apply that to the categories, then anyone can just create a new category when it's been filled up.
> 
> And you're right about indexedDB not being limited, I have a database with millions of items and handling that without issues. It's only an issue if you apply string-based search or filtering on it, fetching keys is no issue at all no matter the size.

I love the idea about adding folders for better organization of conversations!

It could look something like:

image

@tmikaeld Can you open a new issue about this?

tmikaeld:
@Ashesh3

Sweet mockup! It doesn't avoid the huge-DOM issue though, but that's exactly how I'd imagine setting it up.

Maybe it's better if you create it, since I don't want to use your mockup without permission.

Ashesh3:
> @Ashesh3
> 
> Sweet mockup! It doesn't avoid the huge-DOM issue though, but that's exactly how I'd imagine setting it up.
> 
> Maybe it's better if you create it, since I don't want to use your mockup without permission.

Feel free to create the issue, I myself scavenged the mockup it from a template.

Would storing each folder as a separate key (where key is the folder or the date) under the same database, resolve the huge-DOM object issue? In that way, we'd be loading small chunks of conversations, instead of all at once.

tmikaeld:
The problem occurs when you see too much on the screen at the same time.

Something like this solves it:
https://jcoreio.github.io/react-router-drilldown

Since only the category you're in is rendered.

An alternative to handle unlimited amounts is [virtual scroll.](https://devexpress.github.io/devextreme-reactive/react/grid/demos/featured/virtual-scrolling/)

PR #168, opened 2023-10-05
(feat): add support for cloudflare ai gateway openai endpoints

jontybrook:
This PR adds support for Cloudflare AI Gateway endpoints in the API Host field of the advanced OpenAI model configuration.

Cloudflare AI Gateway endpoint URLs are intended to be used as a base path but big-agi expects a host name. I have added a block to detect if a gateway URL is entered in the API Host field (or set via the OPENAI_API_HOST env var) and re-format the host and path accordingly.

This enables simply copy-pasting your cloudflare gateway endpoint from the Cloudflare dashboard into the API Host field.

Before:

screenshot_000445-zqjn2Ugx_05-10-2023_15h39m22s

After: 

screenshot_000446-L2wcQdUg_05-10-2023_15h41m42s

### DIFF
--- a/src/modules/llms/transports/server/openai.router.ts
+++ b/src/modules/llms/transports/server/openai.router.ts
@@ -218,6 +218,31 @@ export function openAIAccess(access: AccessSchema, apiPath: string): { headers: 
   // [Openrouter-specific] lame detection, but works great for now (unless we get requests for non-standard hosts)
   const isOpenrouter = oaiHost.indexOf('openrouter.ai') !== -1;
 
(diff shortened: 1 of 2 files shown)


enricoros:
Thanks for the PR. Do you feel like there should be a separate "Vendor" that has a simple configuration for the cloudflare IDs, or do you prefer the way it's in this PR (i.e. assuming people will use OpenAI > Advanced)

For similar efforts (OpenRouter, LocalAI, Oobabooga, and Azure), we have it as a Vendor.

Also, there's a newer branch which I'll merge Monday and brings some cleanup to the llms module - not a big deal as I can port your PR over.

enricoros:
Update: how can we apply your changes to the new code (in particular the llms framework) that's on `main` ?

PR #240, opened 2023-11-22
fix: ollama listModel endpoint when a model doesn't have TEMPLATE

g1ibby:
fix for https://github.com/enricoros/big-agi/issues/239


enricoros:
Thanks @g1ibby ! really useful. I made the parsers strict to catch if there are changes in expectations form the Ollama vendor. You probably found this fast and saved others ! :)

Issue #282, opened 2023-12-18
Persona Creator


enricoros:
Personas shall be built in multiple manners:

- [x] from unstructured text (plain text, tweets)
- [ ] from structured text (tweets, emails)
- [x] from YouTube Videos
- [ ] from a simple instruction, similarly to the GPTs builder, in ChatGPT
- [ ] ... more ideas?

### Persona From Text
This is different than the **GPTs builder** in ChatGPT. While the GPT builder creates a good system prompt from a short description, this builder is more of a carbon copy of a persona, which will be able to speak/type/behave like the person in the provided text.

While YT persona builder (and improvements in #199) first downloads a transcript and then moved forward with the text from the video, this task's purpose is to skip the first step and only have the user paste text in a text input box and create a persona who could embody that text.

Video from @joriskalz in #287:
![Persona](https://github.com/enricoros/big-AGI/assets/1590910/c30e0ce9-02a6-4f95-bc70-da8d70d2d14c)

PR #287, opened 2023-12-20
Replicate Personas #282

joriskalz:
This feature addresses the need for a more versatile persona creation process as outlined in issue #282. By allowing direct text input, users can quickly and efficiently replicate personas based on a wide range of textual sources.

**Changes made:**
Implemented a tabbed interface in YTPersonaCreator.tsx to allow users to choose between creating a persona from a YouTube video or from pasted text. 

- YTPersonaCreator.tsx: Added imports for tab components, implemented tabbed interface, and added text input handling.
- Updated button states to ensure the 'Create' button is only active when there is input text or a valid YouTube URL.
- PersonaSelector.tsx: Title changed to reflect both use cases.

Looking forward to feedback and suggestions for further refinement.

Screenshots:
Persona

### DIFF
--- a/src/apps/chat/components/persona-selector/PersonaSelector.tsx
+++ b/src/apps/chat/components/persona-selector/PersonaSelector.tsx
@@ -205,8 +205,8 @@ export function PersonaSelector(props: { conversationId: DConversationId, runExa
               <div>
                 <ScienceIcon />
               </div>
-              <div>
-                YouTube persona creator
+              <div style={{ textAlign: 'center' }}>
+                Persona creator
               </div>
             </Button>
           </Grid>}
--- a/src/apps/personas/YTPersonaCreator.tsx
+++ b/src/apps/personas/YTPersonaCreator.tsx
@@ -39,7 +39,7 @@ function useTranscriptFromVideo(videoID: string | null) {
 
 const YouTubePersonaSteps: LLMChainStep[] = [
   {
-    name: 'Analyzing the transcript',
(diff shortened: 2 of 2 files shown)


enricoros:
Thanks @joriskalz your implementation is absolutely excellent!
To celebrate, I will:
- [x] style cleanup using the <TextArea instead of <textarea :D
- [x] rename YTPersonaCreator -> YTCreator
- [ ] remove the "Labs" option, and show this by default
- [ ] merge

Do you think we should default to the **YouTube** tab, or to the **Text** tab?

enricoros:
Thanks @joriskalz - merged, promoted from experimental to stable, and cleaned up some of my code while at that.
Very well! Do you think we should default to the YouTube tab, or to the Text tab?

joriskalz:
> 1. Do you think we should default to the YouTube tab, or to the Text tab?

Thank you for the feedback. I agree that the current implementation is functioning very well and looks great. Regarding the default tab, I believe it might be beneficial to present the YouTube tab first. The YouTube integration showcases a more dynamic and interactive feature that could immediately capture users' interest with its visual appeal.

enricoros:
@joriskalz sounds great. What app did you use to create this beautiful video? I would love to show more of the app features (voice calling, continuous text, diagrams, etc..) using short 15 seconds videos like that.

joriskalz:
@enricoros, i was using [Screen Studio](https://www.screen.studio).

PR #434, opened 2024-02-26
Added Perplexity's new models, updated context lengths, new deprecations

Penagwin:
Perplexity added `sonar-small-chat`, `sonar-medium-chat`, `sonar-small-online`, `sonar-medium-online`

They deprecated these models: `pplx-70b-chat`, `pplx-70b-online`, `llama-2-70b-chat`, and `codellama-34b-instruct`

I wasn't sure how you want to handle deprecations especially with existing chats, so for now I just added a `(deprecated)` tag and changed the descriptions to suggest what to use instead.

> The context window length for several models has been doubled from 8k to 16k, including mixtral-8x7b-instruct and all Perplexity models. 4k tokens are reserved for search results in online models.

I changed all of the current models to 16k context, except for the `online` models. I'm not sure if they mean they only have 4k context, or perhaps it's 12k context and they're saying the output has 4k reserved? I think they might be 12k, as they're using 4k for their bot to parse the results?

https://docs.perplexity.ai/changelog/api-updates-february-2024

Do you want to auto-hide models with "(deprecated)", and perhaps sort them so they show at the bottom of the model selection?


enricoros:
Thansk @Penagwin - solid and merged. Very usefuly.

Regarding the deprecation: I've set the 'hidden' flag on the deprecated models, so that the user will see the models grayed out unless they explicitly unhide them.

Issue #427, opened 2024-02-19
Support for Groq.com


mayurjobanputra:
**Why**
There is a new AI I just discovered called Groq at https://groq.com. I tried it and it performs well but is missing several core features: saving chats and more. 

**Description**
Enable ability to configure a groq API key. See more at https://wow.groq.com/

**Requirements**
Enable ability to configure a groq API key. See more at https://wow.groq.com/

enricoros:
@mayurjobanputra incredible performance, enables new use cases.

I've applied for an API key, but nothing yet. I'd love to add formal support. If you get Access, can you give me a temp key to add the integration until mine comes?

enricoros:
@mayurjobanputra still haven't gotten an API key. Adding support for it shall be trivial.

Honkychunkymonkey:
If you need it, there's https://console.groq.com/docs/quickstart

mayurjobanputra:
I applied. If I do get access, will certainly set you up with a key for testing @enricoros

mayurjobanputra:
This is what they said. So I guess you can get access right away @enricoros 

Enter in your email and the magic link sent to your inbox will provide you access.

enricoros:
Thanks a bunch @mayurjobanputra I'll get to it today and use it for as little as adding support

enricoros:
Closed with @Penagwin's great contribution. Shipping today.

enricoros:
![image](https://github.com/enricoros/big-AGI/assets/32999/761a8a9b-c5d9-4c6c-aa23-423939403896)

PR #438, opened 2024-02-29
Added Groq as an provider

Penagwin:
Adds support for Groq.com

Addresses #427

The logo is a little janky, and heads up that they keep changing their rate limits.
Currently it's
> The rate limit is set to 20 requests per minute and 20 requests per 10 minutes, operating simultaneously. Whichever limit is reached first triggers the rate limit:


enricoros:
Nice! From a quick read it seems you hit all the places to modify to add support - good job.
Will need my laptop to test and merge, will do that soon.

Penagwin:
I discovered Groq does support fetching the models although it's not documented.

So I added support for that

enricoros:
Thanks @Penagwin - great quality contribution. Reviewed and updated the icon.
Groq is really fast, I can consider new use cases with it.
Also good job to find the undocumented model list endpoint - very useful as it reduced ongoing maintenance cost, and shows models as soon as they're released!

PR #507, opened 2024-04-15
Build multi-arch docker image for x64-64 and ARM64

mludvig:
Enable build of ARM64 image for use on MacBook with apple silicon (M1, M2, M3). It's about 3x faster than running the x86-64 container on the Apple silicon under Rosetta emulation.

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -1,3 +1,12 @@
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
+
+# GitHub recommends pinning actions to a commit SHA.
+# To get a newer version, you will need to update the SHA.
+# You can also reference a tag or branch, but the action may change without warning.
+
 name: Create and publish Docker images
 
 on:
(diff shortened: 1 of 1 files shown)


enricoros:
@mludvig does this replace the current images, or adds to them? This workflow is used for the official images in the project page.

mludvig:
@enricoros It builds a multi-arch image. Technically 2 separate images, one for x86-64 and one for arm64 but both have the same name (e.g. big-agi:latest) and the correct platform is pulled based on which arch you're running on.

This won't change the existing images but the next time you release a new version a multi-arch will be created. It should be completely transparent to the users.

My test build is: ghcr.io/mludvig/big-agi:sha-7643770 - you can give it a try.

enricoros:
Thanks, this is very useful and the explanation as well. On the nature of the change itself:
- removing the comments at the beginning?
- changing the versions of many docker images? (I read pinned versions are more secure)

mludvig:
Using "@v5" is a bit more flexible as it may eventually get security or stability updates while being backwards compatible within the v5.x line. These versions were tested and worked for me.

But if you prefer to keep your hashes and put the comments back I can update the PR. Up to you.

mludvig:
@enricoros I've cleaned up the unrelated changes and only left the multi-arch stuff in.

enricoros (approved):

enricoros:
Thanks @mludvig - merged

PR #521, opened 2024-05-04
Docs: fix command to run local build

dandv:


dandv:
Maybe that command should be just `npm start`.

enricoros:
I was thinking so, but 'npm start' doesn't let you specify the port with the same ease, and much instructions are for system integratores.

I've merged your PR and added a note.

PR #533, opened 2024-05-09
Update docker-compose-browserless.yaml

JOduMonT:
- docker made the parameter version is obsolete
- since they are in the same docker-compose it is unnecessary to expose browserless port to access it from big-agi.
- add restart for more conveniency (start when server reboot)

### DIFF
--- a/docs/docker/docker-compose-browserless.yaml
+++ b/docs/docker/docker-compose-browserless.yaml
@@ -8,10 +8,9 @@
 # To start the containers, run:
 #   docker-compose -f docs/docker/docker-compose-browserless.yaml up
 
-version: '3.9'
-
 services:
   big-agi:
+  container_name: big-agi
     image: ghcr.io/enricoros/big-agi:latest
     ports:
       - "3000:3000"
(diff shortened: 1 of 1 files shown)


docs/docker/docker-compose-browserless.yaml
```diff
@@ -22,10 +21,11 @@ services:
     command: [ "next", "start", "-p", "3000" ]
     depends_on:
       - browserless
+    restart: unless-stopped
 
   browserless:
+  container_name: browserless
     image: browserless/chrome:latest
-    ports:
-      - "9222:3000"  # Map host's port 9222 to container's port 3000
```
enricoros, line 29: Is this needed? I see it's removed, but it could be important?

docs/docker/docker-compose-browserless.yaml
```diff
@@ -8,10 +8,9 @@
 # To start the containers, run:
 #   docker-compose -f docs/docker/docker-compose-browserless.yaml up
 
-version: '3.9'
-
 services:
   big-agi:
+  container_name: big-agi
```
enricoros, line 13: I have a lint error here, because the indent is the same as the `big-agi:` and should be more?

enricoros (changes requested):
Thanks, it's not working for me (issues with indent?) and then please let me know if removing the port remap is needed (feels like it was needed to access browserless:3000.

docs/docker/docker-compose-browserless.yaml
enricoros: Note: this does not seem to be needed anymore indeed. do you know if it's safe to remove it for everyone?

PR #575, opened 2024-06-28
[BUG] Fixes text on AI Personas Creator panel

therippa:
Small fix for https://github.com/enricoros/big-AGI/issues/548

image

image


enricoros:
Thanks @therippa ! Is it working? I can't test it right now, but looks like the react hook is missing a dependency (potentially), which means it may not switch text when you switch tab. Is the behavior correct? (Or you can drag the file into big agi and ask :)

therippa:
@enricoros - it's working as expected
big-agi

enricoros (approved):
Reviewed, made small changes to stabilize the react function (dependency was missing in the chain creation hook).

enricoros:
@therippa merged. Performed smaller changes due to non trivial react hook dependencies.
Thanks for the contrib.
https://github.com/enricoros/big-AGI/commit/3a44f70db98b79376c0d718da9c4d71e63fcf207

PR #626, opened 2024-08-23
Update models.data.ts

artoul:


enricoros:
This is now supported alongside o1

PR #646, opened 2024-09-24
Enhanced Gemini API Key Handling (TEST)

darthalex2014:
This fork introduces a more flexible way to handle Gemini API keys, allowing users to specify either a single key or a comma-separated list of keys through the existing GEMINI_API_KEY interface field.

**Current behavior:**
The original code only supports a single Gemini API key.

**Proposed behavior:**
If the GEMINI_API_KEY field contains a single value (no commas), it's treated as a single API key.
If the GEMINI_API_KEY field contains multiple values separated by commas, it's treated as a list of API keys. The application will randomly select one of these keys for each request to the Gemini API.

**Benefits:**
Increased robustness: Using multiple API keys can help prevent rate limiting issues and ensure uninterrupted access to the Gemini API.


enricoros:
Great idea @darthalex2014 - this should be a thing

Issue #600, opened 2024-07-20
[BUG] Voice Feature Not Working on iOS


echohuiecho:
### Description

Hello,

I encountered an issue with the voice feature on iOS. I tried deploying the project using Vercel, and it seems that the sound cannot be played automatically in either the voice call UI or the messages UI. I have verified that the request was successfully sent to ElevenLabs.

**Steps to Reproduce:**
1. Deploy the project using Vercel.
2. Access the application on an iOS device using Chrome or Safari
3. Set the ElevenLabs API key
4. Inside Preferences > Voice > Speak Responses, set as 'Start' or 'Full'
5. Attempt to send messages in the voice call UI or messages UI and wait for the response to be played by ElevenLabs voice

**Expected Behavior:**
The sound should play automatically when the message is sent in both the voice call UI and messages UI.

**Actual Behavior:**
The sound does not play automatically in either the voice call UI or messages UI.

**Additional Information:**
- The request to ElevenLabs was successfully sent.

### Device and browser

- Affected devices: iOS
- Browsers tested: Chrome and Safari on iOS

### Willingness to Contribute

- [ ] 🙋‍♂️ Yes, I would like to contribute a fix.

matrix303:
Hi @enricoros, reviving a old bug but I faced the same problem, both on chrome and safari on iOS.

Looking into whats happening, it seems that safari is throwing the following error (attached). I checked the github and it seems that MediaSource is being called in the  LiveAudioPlayer util and may not be initializing there properly.

Note: I confirmed that "MediaSource" feature flags are enabled in the iOS settings.

<img width="688" alt="Image" src="https://github.com/user-attachments/assets/a2f1e156-08b5-4507-b012-c21c8adbebda" />

<img width="1098" alt="Image" src="https://github.com/user-attachments/assets/87f8b4a6-1cbe-4395-98a8-1fd867c4b265" />

enricoros:
Retitle to the residual: verify iOS playback after the MediaSource guard. @matrix303 the class you traced is gone - playback now feature-checks MediaSource and falls back to buffered playback (AudioLivePlayer.isSupported / AudioAutoPlayer), so the crash you screenshotted cannot happen on 2.1. If audio is still silent on iOS, the likely remaining cause is the autoplay gesture policy, which is a different fix; a fresh confirmation on 2.1 would settle it.

PR #661, opened 2024-10-17
Support Web Speech API

zoollcar:
New feature:
image

## TODO:
* Some languages ​​have very few voices. pre-select will causing the voice list to become empty. Perhaps we should hide the pre-select option in these cases.
* When switching the TTS engine, the Elecenlabs test voice doesn't stop when I switch to the Web Speech API.
* The upstream project has a some [broken json files](https://github.com/HadrienGardeur/web-speech-recommended-voices/blob/abc38af340853335cf6e7c6b223f79c7c5a46eac/json/kn.json#L97), I'll fix and make a PR there

module has been moved to modules/browser/speech-synthesis

> The Web Speech API test sentence is from the web-speech-recommended-voices project and contains a placeholder {name}. 

{name  }has been replaced to the name of voice

> Using a Git submodule might not be the best solution.

copy en.json as a local file [Languages.json](https://github.com/zoollcar/big-AGI/blob/Browser-TTS-for-version2/src/modules/browser/speech-synthesis/preSelect/Languages.json). fetch voice list when select a language


enricoros:
Thanks for the feature. This patch is now in a state where I can review it and potentially merge it.

enricoros:
Update: thanks for resubmitting the PR, this is definitely a higher quality code that considers the application (e.g. other modules).

I'm testing it on mobile and it's hanging a couple of times (I believe it to be a stability error with some changing react reference) and it's possibly something I can fix, but it's gonna require some time for me to check out and develop.

On the UX side, there could be some rough edges (on my android phone the High quality List doesn't do much, no matter what one chooses the experience doesn't change, and this happens for the 4 available voices as well). So there's something that I can look into to improve the UX. Why is key? Because every feature Big AGI gets the same scrutiny and UX perfection.

Thanks again, I'll follow up when I have time to check this out and review and change what needs to be changed. Let me know in the meantime if anything can improve on your side.

src/common/components/useVoiceCapabilities.ts
```diff
@@ -0,0 +1,74 @@
...
+import { EXPERIMENTAL_speakTextStream as EXPERIMENTAL_elevenlabsSpeakTextStream } from '~/modules/elevenlabs/elevenlabs.client'
+
+import { useCapability as useBrowserSpeechSynthesisCapability } from '~/modules/browser/speech-synthesis/browser.speechSynthesis.client'
+import { speakText as browserSpeechSynthesisSpeakText } from '~/modules/browser/speech-synthesis/browser.speechSynthesis.client'
+import { EXPERIMENTAL_speakTextStream as EXPERIMENTAL_browserSpeechSynthesisSpeakTextStream } from '~/modules/browser/speech-synthesis/browser.speechSynthesis.client'
+
+import { useElevenLabsVoices } from '~/modules/elevenlabs/useElevenLabsVoiceDropdown';
+import { useBrowserSpeechVoices } from '~/modules/browser/speech-synthesis/useBrowserSpeechVoiceDropdown';
+
+export const TTSEngineList: string[] = [
```
enricoros, line 15: See other note on this. I wonder if this would be better served as a map:
```typescript
export type TTSEngineKey = 'elevenlabs' | 'webspeech';

const TTSEngineList: { [key in TTSEngineKey]: { label: string, priority: number } } = {
  'elevenlabs': {
    label: 'ElevenLabs',
    priority: 2,
  },
  'webspeech': {
    label: 'Web Speech API',
    priority: 1,
  },
};
```
I've added an attribute called 'priority' to show how we can extend it in the future, for instance with default configurations.

src/common/components/useVoiceCapabilities.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export function useCapability(): CapabilitySpeechSynthesis {
+  return getConditionalCapability()();
+}
+
+
+export async function speakText(text: string, voiceId?: string) {
+  const TTSEngine = getTTSEngine();
+  if (TTSEngine === 'Elevenlabs') {
+    return await elevenlabsSpeakText(text, voiceId);
```
enricoros, line 57: This is a way of doing it. I wonder if we should have a common Interface, and have all the TTS providers (e.g. OpenAI has 2 as well, one via their TTS and one in the new [audio] models), and there's Play.ht and more.

A way to reduce the switch-cases (or ifs) will be to have a common interface, such as ISpeechSynthesis, that will be returned by getTTSEngine (basically instead of the string it would return an object that implements the interface).

Just a thought.

src/modules/browser/speech-synthesis/preSelect/Languages.json
```diff
@@ -0,0 +1,75 @@
+{
+  "ar": "Arabic",
```
enricoros, line 2: For the languages we also have another list of TTS/ASR Languages in the config. I wonder if we need 2.

src/modules/browser/speech-synthesis/BrowserSpeechSettings.tsx
```diff
@@ -0,0 +1,111 @@
...
+  const [testUtterance, setTestUtterance] = React.useState<string | null>(null);
+  const [voiceNameFilters, setVoiceNameFilters] = React.useState<string[] | null>(null);
+
+  // external state
+  const [languageCode, setLanguageCode] = useLanguageCodeForFilter();
+
+  React.useEffect(() => {
+    if (languageCode) {
+      const fetchFunction = async () => {
+        let res = await fetch(`https://raw.githubusercontent.com/HadrienGardeur/web-speech-recommended-voices/refs/heads/main/json/${languageCode}.json`);
```
enricoros, line 26: Well done here.

src/modules/browser/speech-synthesis/BrowserSpeechSettings.tsx
```diff
@@ -0,0 +1,111 @@
...
+        voices = voices.filter((voice: any) => {
+          return voice.quality.includes('high') || voice.quality.includes('veryHigh');
+        });
+        let voiceNameFilters = voices.map((voice: any) => voice.name);
+        setTestUtterance(data.testUtterance);
+        setVoiceNameFilters(voiceNameFilters);
+      };
+      fetchFunction().catch((err) => {
+        console.log('Error getting voice list: ', err);
+        addSnackbar({ key: 'browser-speech-synthesis', message: 'Error getting voice list', type: 'issue' });
```
enricoros, line 38: I got this message with some of the languages of the list. Strange because I thought the list will have all valid languages.

src/modules/browser/speech-synthesis/store-module-browser.tsx
```diff
@@ -0,0 +1,40 @@
+import { create } from 'zustand';
+import { persist } from 'zustand/middleware';
+import { useShallow } from 'zustand/react/shallow';
+
+export type BrowsePageTransform = 'html' | 'text' | 'markdown';
```
enricoros, line 5: This probably doesn't belong here. We have already a browser store (for the browsing capability) but it's different.

src/modules/browser/speech-synthesis/store-module-browser.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+interface BrowseState {
+
```
enricoros, line 8: I'm wondering if the settings also of TTSEngine could be here, to keep everything together.

src/modules/browser/speech-synthesis/useBrowserSpeechVoiceDropdown.tsx
```diff
@@ -0,0 +1,124 @@
+import * as React from 'react';
+
+import { CircularProgress, Option, Select } from '@mui/joy';
+import KeyboardArrowDownIcon from '@mui/icons-material/KeyboardArrowDown';
+import RecordVoiceOverTwoToneIcon from '@mui/icons-material/RecordVoiceOverTwoTone';
+
+import { useBrowseVoiceId } from './store-module-browser';
+import { speakText, cancel } from './browser.speechSynthesis.client';
+
+function VoicesDropdown(props: {
```
enricoros, line 10: I see this is a duplication of the ElevenLabs. probalby needed because of different logic.

enricoros (changes requested):
Overall there could be some details (and a crash) to be ironed out. Going in the right direction and increasing in quality.

I wonder if the whole module should get a cleanup, meaning to bring also elevenlabs and webspeech under a same umbrella, e.g. /modules/tts/* or similar. 

The module would benefit likely from having an abstraction (e.g. interface ISpeechSynthesis or similar) with the 2 Engines implementing the same interface. This way every call becomes more abstract and the caller doesn't need to know "if elevenlabs do this, otherwise do that".

In general is good, please take a look at the comments (only come are necessary).

src/common/components/useVoiceCapabilities.ts
zoollcar: After a try. I think use map for extending is a good idea, but i suggest use a list contained map:

```ts
export type TTSEngineKey = 'Elevenlabs' | 'Web Speech API';
export type ASREngineKey = 'Web Speech API';

export const TTSEngineList: { key: TTSEngineKey; label: string }[] = [
  {
    key: 'Elevenlabs',
    label: 'ElevenLabs',
  },
  {
    key: 'Web Speech API',
    label: 'Web Speech API',
  },
];
```

because there are only 2 case using the variable now, one is `<FormRadioControl>` in `VoiceSettings.tsx` another one is `store-app-chat.ts`. `<FormRadioControl>` need Iterating and `VoiceSettings.tsx` need a default value. These are map can't do.

src/modules/browser/speech-synthesis/preSelect/Languages.json
zoollcar: Languages is a big problem even we have ISO languages code. People don't use it latest version ant it's chaos itself. so the reality is different provider have different code.

There are some solution:
1. For different modals/provider, use different languages file.
 benefit: easy to update, save dev energy, have more choices even some small languages
 issues: hard to link different part of app, user need to setting one by one.
2. use one languages file, manual maintenance different system's code
benefit: user set once. setting everywhere
issues: can only maintain intersection part, less languages. need manually maintain. Introducing additional complexity so may have bugs.

src/modules/browser/speech-synthesis/BrowserSpeechSettings.tsx
zoollcar: Upstream error, the language listed in json file does not have a corresponding file. so 404. I'll delete all invalid languages

zoollcar:
I'll make a abstraction(under `modules/tts/`, `ISpeechSynthesis`). The current plan is to refer to the llms module. 

The refactored version will be updated these days.

zoollcar:
Basically done the abstraction. What is worth mentioning is a change of UI:
image
Engine selection change to drop-down box for more options and mobile compatibility.

michieal:
Okay, so... from my initial tests, this PR is spot on!
I did notice though, that it doesn't work in offline mode. This raises a couple of questions for me; not a bad thing mind you, just curiosity.
First - is the output text sent out to become speech?
Second - is there a way to have it download the voice files and process it locally? I am asking about this, as one of the reasons to use a localized AI is that people don't want everything going out across the internet. So, on that thought, maybe a notice in the Voice area stating that it requires an active connection would be a good idea.

All in all - quite impressed with this! It has everything that I put into the Feature Request that I made (#703) so, you rock! :)

enricoros:
@michieal thanks for reviewing and testing this. It's a great change that shall come in. Quick replies:
- I've pushed 7626b48b which shall remove the flooding of "[DEV] AIX: OpenAI-dispatch missing completion...". This was probably because the local models didn't return the correct completion tokens (0 when not zero). I've silenced the message and accept it to be 0
- Do you have more details on the localAI issue
- You're correct on ollama not working because this PR is not on the tip of `v2-dev` but earlier
- Great point on asterisks stripping - there should be some de-markdown'ification of the text before being sent to any TTS. Will try to remember this

Huge thanks for @zoollcar for this PR, as @michieal says it's tested and solid. Let us know if there are recent changes.

michieal:
> @michieal thanks for reviewing and testing this. It's a great change that shall come in. Quick replies:
> 
> * I've pushed [7626b48](https://github.com/enricoros/big-AGI/commit/7626b48b70f0c3c274704d43f9658a12700f8670) which shall remove the flooding of "[DEV] AIX: OpenAI-dispatch missing completion...". This was probably because the local models didn't return the correct completion tokens (0 when not zero). I've silenced the message and accept it to be 0

Thank you!! big help!

> * Do you have more details on the localAI issue

Apparently, it's a LocalAI issue with its backends on the latest release. I routed the output of it to a log file and then matched up timestamps. Double checked with the V2-dev version that I have that worked, and the issue is still there... so yeah, not Big-AGI's fault. (It's a GRPC issue... "failed to load backend".)

Happy to help out with this. Sorry that I am not able to help out on the coding side. :)

matrix303:
Tried out the PR, fantastic work @zoollcar and @enricoros, it did require some finagling with merging with the latest commits on v2-dev as of Jan 30, 2025. Overall very high quality code and cant wait to be merged to have a alternative for the iOS call issues mentioned in #600 while the actual error with elevenLabs on iOS gets worked out.

enricoros:
Thank you for giving it a try @matrix303 -- Was there lots of work to port it to the current v2-dev? I also have a more experimental branch that can't be released yet and I know there will be conflicts, I hope only few.
This code seems high quality now.

matrix303:
Honestly, I got most of it working in two or three hours. My main problem was the unfamiliarity with how the code has changed since the last PR commit and your v2-dev changes and thus took longer than it should have.

For some reason I had issues with the following, but I think unrelated:
- Chat drawer
- when you start a call, there's a compatibility checklist page, that was still trying to detect for elevenlabs API setup,but if can be avoided if web speech, API is the new default?
- for the voices: You had to select the language first, which would filter all available voices and only show the correct one. Selecting that works. If I directly go and select a voice without for selecting a language, then some bugginess there. But some conditional logic should resolve it.

enricoros:
This is a great heads up. The hard part in this case is to come up with the perfect framework that can allow to apply different engines in the house automated detection and fallbacks.

Something like: with 0 configuration works already, and with the best language for your chrome / Safari / android/ iphone, etc. 

If then you configure a service, which includes the OpenAI TTS voices (around 40% of Big-AGI users use OpenAI, I feel) and Play.ht, ElevanLabs, etc. then the framework should use either the highest priority service or fall back to Web Speech (again with the best preset).

Big challenge. And also, I haven't looked into the LocalAI interface for TTS, but it's another way to have a solution.

matrix303:
Yeah! You make excellent points and I think a framework that lets you enable multiple remotes, localAI and web speech would be excellent.

I think the abstraction in this PR is a great first step that works to setup this unified framework.

enricoros:
@zoollcar @matrix303 @michieal Official support for WebSpeech has been added. Multiple providers including Web Speech (free), LocalAI (free), OpenAI, ElevenLabs can now be configured and used for various services.

PR #716, opened 2025-01-07
feat: add custom API endpoint support for Google Gemini

Jonathan523:
add custom API endpoint support for Google Gemini
fix the API key management URL

### DIFF
--- a/src/modules/llms/vendors/gemini/GeminiServiceSetup.tsx
+++ b/src/modules/llms/vendors/gemini/GeminiServiceSetup.tsx
@@ -35,7 +35,7 @@ const SAFETY_OPTIONS: { value: GeminiWire_Safety.HarmBlockThreshold, label: stri
 export function GeminiServiceSetup(props: { serviceId: DModelsServiceId }) {
 
   // advanced mode
-  const advanced = useToggleableBoolean(true);
+  const advanced = useToggleableBoolean(false);
   // external state
   const { service, serviceAccess, serviceHasBackendCap, serviceHasLLMs, serviceSetupValid, updateSettings } =
     useServiceSetup(props.serviceId, ModelVendorGemini);
--- a/src/modules/llms/vendors/gemini/GeminiServiceSetup.tsx
+++ b/src/modules/llms/vendors/gemini/GeminiServiceSetup.tsx
@@ -17,9 +17,11 @@ import { useLlmUpdateModels } from '../../llm.client.hooks';
 import { useServiceSetup } from '../useServiceSetup';
 
 import { ModelVendorGemini } from './gemini.vendor';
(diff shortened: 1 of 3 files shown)


src/modules/llms/vendors/gemini/GeminiServiceSetup.tsx
```diff
@@ -17,9 +17,11 @@ import { useLlmUpdateModels } from '../../llm.client.hooks';
 import { useServiceSetup } from '../useServiceSetup';
 
 import { ModelVendorGemini } from './gemini.vendor';
+import { useToggleableBoolean } from '~/common/util/hooks/useToggleableBoolean';
+import { FormTextField } from '~/common/components/forms/FormTextField';
 
 
-const GEMINI_API_KEY_LINK = 'https://makersuite.google.com/app/apikey';
+const GEMINI_API_KEY_LINK = 'https://aistudio.google.com/app/apikey';
```
enricoros, line 24: Thanks, this is great too

src/modules/llms/vendors/gemini/GeminiServiceSetup.tsx
```diff
@@ -90,7 +94,15 @@ export function GeminiServiceSetup(props: { serviceId: DModelsServiceId }) {
       {/*of being unsafe.*/}
     </FormHelperText>
 
-    <SetupFormRefetchButton refetch={refetch} disabled={!shallFetchSucceed || isFetching} loading={isFetching} error={isError} />
+    {advanced.on && <FormTextField
+      autoCompleteId='openai-host'
```
enricoros, line 98: smaller change here, I'll rename to 'gemini-host'

enricoros (approved):
All good, I'll merge.

norrybul:
I'm just new to big-AGI. I hope this is related to this thread. Setting my Gemini API key doesn't work (Bad Request error) possibly due to geolocation. It was alright before on other UIs as I remembered. But lately it doesn't, and also on my first try of big-AGI.
image
Using a proxy does the trick, since there's no API endpoint input option - I wish there was like OpenAI!
image
With reference to this Gemini link: https://ai.google.dev/gemini-api/docs/openai , I was able to avoid using proxies using their compatible endpoints (https://generativelanguage.googleapis.com/v1beta/openai/...) as a workaround. I hope it helps others like me. Thanks.

Jonathan523:
@norrybul The reason you don't have the API endpoint input option is that the official  deployed version of https://get.big-agi.com is NOT the latest version（it actually uses the `v1-stable` branch and the feature is on branch `v2-dev`). Instead, you need to use versions deployed with `v2-dev` branch. You can do this by deploying your own version on vercel.

PR #719, opened 2025-01-08
fix(gemini): fix FormHelperText for safety settings visibility 

Jonathan523:
fix(gemini): make FormHelperText for safety settings invisible when advanced is not activated


enricoros:
Thanks @Jonathan523, merged.

PR #724, opened 2025-01-15
ChatDrawer: virtualize scroll (for very numerous chats)

enricoros:


enricoros:
Note: closing this because we have overly optimized the Chat Drawer list now, especially on app.big-agi.com

PR #771, opened 2025-03-10
fix: detect mobile for newline behavior

dogmatic69:
if mobile will remain as is, if not mobile will set to undefined which will return desktop to expected functionailty for return vs shift return

fixes enricoros/big-AGI#770


dogmatic69:
tested and confirmed this makes it work as expected

Issue #744, opened 2025-01-29
[BUG] Unable to enum o1 models on Azure OpenAI


alkampfergit:
### Environment

big-agi.com (production website)

### Description

When I try to refresh my deployment, if I deploy an o1-mini model I got this error

[ { "received": "failed", "code": "invalid_enum_value", "options": [ "succeeded" ], "path": [ "data", 3, "status" ], "message": "Invalid enum value. Expected 'succeeded', received 'failed'" } ]

I've deployed o1-mini on another datacenter, this time I do not get error, but the model does not show in the list of models.

### Device and browser

WIndows, edge browser.

### Screenshots and more

![Image](https://github.com/user-attachments/assets/cf8bbe4b-0dcc-4824-8205-8df2e1e45175)

### Willingness to Contribute

- [ ] 🙋‍♂️ Yes, I would like to contribute a fix.

enricoros:
Hi @alkampfergit can you try now? I've just relaxed the parser and should proceed.

alkampfergit:
I'll upgrade and try

Il mer 29 gen 2025, 19:40 Enrico Ros ***@***.***> ha scritto:

alkampfergit:
I tried to update docker image but I didn't see any new image, still have the same problem

rafaelpierre:
same here, I don't get any error messages but my o1 endpoints from Azure don't show up

enricoros:
Reopening because evidently there's an issue with Azure models. Note: the complexity with azure it's always been the dynamic listing, as the names of deployments are likely non standard and one cannot guess the context window sizes, model capabilities, etc. The code right now is looking up the official OpenAI model names, but in this particular instance there could be other underlying issues.

rafaelpierre:
Wouldnt make sense to ask users to include the deployment name in the URL
for Azure models? That would solve the issue.

enricoros:
@rafaelpierre it does, but everything else is auto detected. I welcome any pull requests as I don't have azure deployments so it would take me longer to test. 1. For the missing o1, I'll take a look. 2. For the missing UI, I welcome anyone's PR. Note: one must configure also the context window, and capabilities (image input, function calling, etc.) and QUIRKS of the model. For instance, o1 can't stream, and the user will need to say that (again, no auto detect with azure models) otherwise the LLM call will just fail (and note that O1 and P1 preview and O3 have all different quirks)

enricoros:
@rafaelpierre all models are now enumerated correctly

[...]

PR #774, opened 2025-03-13
fix: get o1 working

dogmatic69:
max_completion_tokens is the new way, I believe it works fine for the others.

Perhaps "api-version" wants to be a config, I just used the latest.

### DIFF
--- a/src/modules/aix/server/dispatch/chatGenerate/adapters/openai.chatCompletions.ts
+++ b/src/modules/aix/server/dispatch/chatGenerate/adapters/openai.chatCompletions.ts
@@ -66,6 +66,9 @@ export function aixToOpenAIChatCompletions(openAIDialect: OpenAIDialects, model:
   if (hotFixAlternateUserAssistantRoles)
     chatMessages = _fixAlternateUserAssistantRoles(chatMessages);
 
+  if (model.id === 'o1' || model.id.startsWith('o1-'))
+    model.temperature = null; // NOTE: o1 models don't support temperature, so we set it to null
+
 
   // Construct the request payload
   let payload: TRequest = {
@@ -74,7 +77,7 @@ export function aixToOpenAIChatCompletions(openAIDialect: OpenAIDialects, model:
     tools: chatGenerate.tools && _toOpenAITools(chatGenerate.tools),
     tool_choice: chatGenerate.toolsPolicy && _toOpenAIToolChoice(openAIDialect, chatGenerate.toolsPolicy),
     parallel_tool_calls: undefined,
-    max_tokens: model.maxTokens !== undefined ? model.maxTokens : undefined,
+    max_completion_tokens: model.maxTokens !== undefined ? model.maxTokens : undefined,
     ...(model.temperature !== null ? { temperature: model.temperature !== undefined ? model.temperature : undefined } : {}),
     top_p: undefined,
     n: hotFixOnlySupportN1 ? undefined : 0, // NOTE: we choose to not support this at the API level - most downstram ecosystem supports 1 only, which is the default
--- a/src/modules/llms/server/openai/openai.router.ts
+++ b/src/modules/llms/server/openai/openai.router.ts
@@ -96,7 +96,7 @@ export const llmOpenAIRouter = createTRPCRouter({
 
       // [Azure]: use an older 'deployments' API to enumerate the models, and a modified OpenAI id to description mapping
       if (access.dialect === 'azure') {
-        const azureModels = await openaiGETOrThrow(access, `/openai/deployments?api-version=2023-03-15-preview`);
+        const azureModels = await openaiGETOrThrow(access, `/openai/deployments?api-version=2025-02-01-preview`);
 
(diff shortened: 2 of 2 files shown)


dogmatic69:
Probably solves #744

enricoros:
Hi @dogmatic69, can you tell me more about what this fixes and how to test it? o1 has been working for me for long time.

dogmatic69:
Without the following, o1 is not even listed in the available models for azure 
```
          .filter(m => m.model.includes('gpt') || m.model === 'o1')
```

Once it was shown, There are some errors which the rest fixes. o1 required a newer API version, temperature is not supported by o1, the other field is depreciated for max tokens.

src/modules/aix/server/dispatch/chatGenerate/adapters/openai.chatCompletions.ts
```diff
@@ -74,7 +77,7 @@ export function aixToOpenAIChatCompletions(openAIDialect: OpenAIDialects, model:
     tools: chatGenerate.tools && _toOpenAITools(chatGenerate.tools),
     tool_choice: chatGenerate.toolsPolicy && _toOpenAIToolChoice(openAIDialect, chatGenerate.toolsPolicy),
     parallel_tool_calls: undefined,
-    max_tokens: model.maxTokens !== undefined ? model.maxTokens : undefined,
+    max_completion_tokens: model.maxTokens !== undefined ? model.maxTokens : undefined,
```
dogmatic69, line 80: Look at the source for `TRequest`, that is where it says depreciated

src/modules/llms/server/openai/openai.router.ts
```diff
@@ -114,7 +114,7 @@ export const llmOpenAIRouter = createTRPCRouter({
 
         // only take 'gpt' models
         models = azureWireModels
-          .filter(m => m.model.includes('gpt'))
+          .filter(m => m.model.includes('gpt') || m.model === 'o1')
```
dogmatic69, line 117: without this line, the model does not even show up in the options

src/modules/llms/server/openai/openai.router.ts
```diff
@@ -96,7 +96,7 @@ export const llmOpenAIRouter = createTRPCRouter({
 
       // [Azure]: use an older 'deployments' API to enumerate the models, and a modified OpenAI id to description mapping
       if (access.dialect === 'azure') {
-        const azureModels = await openaiGETOrThrow(access, `/openai/deployments?api-version=2023-03-15-preview`);
+        const azureModels = await openaiGETOrThrow(access, `/openai/deployments?api-version=2025-02-01-preview`);
```
dogmatic69, line 99: image

src/modules/aix/server/dispatch/chatGenerate/adapters/openai.chatCompletions.ts
dogmatic69: also as per docs 
image

dogmatic69:
you say o1 has been working, but in azure? I highly doubt that as the API version for a start 100% did not support o1.

enricoros:
@dogmatic69 you're right, I was talking about o1, not the Azure o1.

The line you mention is def needed. And how about removing the full filter, as then you have 'o3', etc.

I'm more worried about the max_tokens change. Looks like it would apply to All openai dialects (we support ~10 OpenAI derivatives), is it the case?

I can't tell if it would break any other OpenAI-like provider but definitely it looks like so. And even for OpenAI models, not all of them were transitioned to the new Max Competition tokens.

Let me know if there are pieces of this patch that are safe and want me to merge.

src/modules/llms/server/openai/openai.router.ts
enricoros: I've removed this in my version.

src/modules/llms/server/openai/openai.router.ts
```diff
@@ -391,7 +391,7 @@ export function openAIAccess(access: OpenAIAccessSchema, modelRefId: string | nu
       if (apiPath.startsWith('/v1/')) {
         if (!modelRefId)
           throw new Error('Azure OpenAI API needs a deployment id');
-        url += `/openai/deployments/${modelRefId}/${apiPath.replace('/v1/', '')}?api-version=2023-07-01-preview`;
+        url += `/openai/deployments/${modelRefId}/${apiPath.replace('/v1/', '')}?api-version=2025-02-01-preview`;
```
enricoros, line 394: This is an insightful change to make those OpenAI Deployments work. Applied.

enricoros (commented):
Thanks for the tests and ideas.

enricoros:
@dogmatic69 Azure is working great with the 4 commits above.

Models list from my deployments to East US 2:
image

And all models being used, o3, o1, 4o, with/without image support:
image

This is closed now.

enricoros:
o1 fully works, including advanced parameters, such as "reasoning effort" or "restore markdown":
image

dogmatic69:
yes great, works now. Thanks 🥇

PR #776, opened 2025-03-15
AIX: AI: multi-key support.

darthalex2014:
ALIBABA_API_KEY
DEEPSEEK_API_KEY
GROQ_API_KEY
MISTRAL_API_KEY
PERPLEXITY_API_KEY
TOGETHERAI_API_KEY
XAI_API_KEY


enricoros:
@darthalex2014 I love it. Since it's a single file, can you have a small function (like function anyKey(multiKeys: string) { ... }) and then use that function in all places?

Single place instead of 5 :) easier to maintain and check for security.

PR #779, opened 2025-03-20
Unfinished PR to enable built in TTS in voice call

pydlv:


enricoros:
Thanks for the code @pydlv. To make it easy to adopt or continue it, can you document with a couple of bullets:

- [ ] what's your architecture idea
- [ ] what's done
- [ ] what's to do
- [ ] any other thoughts :)

enricoros:
@pydlv Done. Multi-provider TTS is now available in Big-AGI Open

PR #798, opened 2025-04-25
Personas creation & saving to localStorage

duzinhoml:
- User can create custom AI persona with: title, description, system prompt, symbol, and examples
- User-created personas are saved via localStorage and persist


enricoros:
Hi @duzinhoml, great work. I cannot merge this to the main `v2-dev` branch because it conflicts with the new Personas management and Sync system we have built in the version on app.big-agi.com.

Thanks for contributing it!

This PR will stay open and is for anyone that wants to enable creation and saving of Personas on localStorage.

enricoros:
Closing.

I kept this open as a fork reference, but 16 months of drift settled it: personas moved to the synced entity system, localStorage is a direction we're leaving, and the files this touches keep moving. There is no future point where this becomes mergeable.

Thanks for the contribution - the creation form was a nice reference. This close comes from a PR-backlog pass; if you disagree, comment and I'll reopen.

PR #806, opened 2025-05-24
Fix: Disable pane closure buttons when >2 panes to prevent crash

datainvest23:
The application was crashing when you attempted to close panes via the 'Close Tab' or 'Close Other Tabs' buttons in the `PaneTitleOverlay` component if more than two panes were open.

This was due to a known limitation or assertion issue within the `react-resizable-panels` library when panes are removed in layouts with more than two active panes.

This commit addresses the issue by disabling these buttons in `PaneTitleOverlay.tsx` when the `numberOfPanes` (derived from `chatPanes.length` via the `usePanesManager` hook) is greater than 2.

This prevents the crash and ensures application stability, albeit with a temporary functional limitation on closing panes in complex multi-pane views until the underlying library issue can be better addressed.


enricoros:
Hi @datainvest23, I believe this issue was fixed since this PR. We have upgraded to newer react-resizeable-panels and the issue cannot be reproduced.

PR #811, opened 2025-06-05
Add Claude 4 thinking variants via OpenRouter

sam0jones0:
closes: https://github.com/enricoros/big-AGI/issues/807

Adds support for Claude 4's thinking feature via OpenRouter.

- Creates duplicate model entries in dropdown for thinking variants (e.g., "Claude Sonnet 4" and "Claude Sonnet 4 (thinking)")
- Implements thinking parameter support with configurable budget
- Adds debug logging for tracking thinking budget values

Screenshot 2025-06-05 at 17 29 22


enricoros:
I've been reviewing this but it's taking longer than expected. There are some instances of deleted code (e.g. from the non streaming parser) which I don't know why... the most likely explanation is that AI has decided to remove that code, but that's wrong. I fear about side effects that will happen to other providers.

enricoros:
@sam0jones0 I've cleaned this up - can you give `v2-dev` a spin and check if this cleaned up version works perfectly?

PR #830, opened 2025-08-14
V2 dev paulshort

powyncify:


enricoros:
Fully merged. Thanks @paulhshort. Please @powyncify, @paulhshort let me know if it's working as before, or better than before. Hopefully there aren't issues and the code will be more maintainable as well.

powyncify:
All resolved. Thank you both. Onwards!

PR #838, opened 2025-09-30
feat: Add conditional dropdowns for Verbosity and Reasoning Effort if model is supported

HarryVu176:
Some OpenAI models support Verbosity / Reasoning Effort parameters. I think it will be handy to be able to change those settings on the toolbar for quick (especially reasoning effort) instead of having to go to model dropdown -> model settings.


HarryVu176:
image
I have attached the image here

enricoros:
Hi @HarryVu176 - this is a great change and well done. I can se why it's convenient to have params there.
Please take a look at https://app.big-agi.com, which I've just dark launched, and has parameters setting in the convenient side-pane. May that be useful to you, or do you prefer in the dropdowns?

HarryVu176:
> Hi @HarryVu176 - this is a great change and well done. I can se why it's convenient to have params there. Please take a look at https://app.big-agi.com, which I've just dark launched, and has parameters setting in the convenient side-pane. May that be useful to you, or do you prefer in the dropdowns?

Thanks a lot for checking the PR! I'm no UI expert, I tried agi.big-agi.com and it looks polished. On my vertical monitor/small screen, I rarely toggle the right panel, I usually leave it closed, so having params in the dropdowns works well. I think it also depends on how often people change verbosity/reasoning, I tend to switch between depending on my question/context to save tokens.
That's what I thought, but I will follow your design guidelines and decisions!

DoS007:
Easy switching for reasoning or websearch is relevant (didn't test new github repo changes).

_Include reasoning and websearch in message author name_: I also think, that knowing which answer was given with how much thought or sources (reasoning or websearches, but especially reasoning as you cannot see it as directly) is relevant.

Btw. when using Beam and take all messages without combination it would be nice to see each answer better or see which models were used instead of "unk-model". And i believe when the first message is beam the chat name doesn't get genereted automatically.

tlima:
I love this idea @HarryVu176. Would be nice to place the temperature control there as well.

enricoros:
@tlima there are quite many controls that could be there: search controls, reasoning effort, reasoning tokens, image generation on/off, and more and more. In the app.big-agi.com I put the controls on the right panel, quickly accessible, and probably that's the best pattern, as the top bar will be getting crowded quickly, and even worse on Mobile, where the top bar is already too crowded with Persona and Model selection.
I haven't back ported the changes from the app.big-agi.com to the side panel of this main branch yet, as it requires significant time (1-2 days).
If someone wants to maintain a branch with the controls on the top bar, I support it.

enricoros:
Closing this one after a long soak - the direction we settled on is quick parameter access in the side panel (as shipped on app.big-agi.com) rather than toolbar dropdowns, and the toolbar surface has since diverged quite a bit. Thanks for the well-made contribution and the discussion it sparked - the convenience need it identified is real and is being served by the panel approach.

PR #985, opened 2026-02-18
better system prompts

DigitallyTailored:


DigitallyTailored:
Ignore this - wrong direction!

PR #1009, opened 2026-03-04
feat(openai): add HTTP-Referer header to identify big-AGI traffic

MarkosBK:
Adds `HTTP-Referer: https://big-agi.com` to outgoing OpenAI-compatible API requests.

**Why:** Several API providers (notably LLMAPI) use the `HTTP-Referer` header 
to attribute traffic in their analytics dashboards. This makes big-AGI requests 
identifiable, which helps both providers and users tracking their usage.

Single-line change using the existing `BaseProduct.ProductURL` constant.


enricoros:
Need to regression check this with all OpenAI-compatible providers to make sure it doesn't stop/prevent others from working.

MarkosBK:
> Need to regression check this with all OpenAI-compatible providers to make sure it doesn't stop/prevent others from working.

How do I do that?

enricoros:
@MarkosBK It's prob something I have to do because I have the accounts to a dozen openai-compatible providers. My plate is a bit full tho.

MarkosBK:
> @MarkosBK It's prob something I have to do because I have the accounts to a dozen openai-compatible providers. My plate is a bit full tho.

Thank you very much, it would be really cool to be able to track how many users are using the provider specifically from your platform.

PR #1011, opened 2026-03-04
store-logic-sherpa: guard usage count increment against SSR

Blacksuan19:
The useLogicSherpaStore.setState() call at module level ran during server-side rendering where localStorage is unavailable, causing a hydration crash. Wrap with isBrowser so it only executes in the browser context.


enricoros:
Thanks @Blacksuan19 - good fix. merged.

PR #1012, opened 2026-03-04
store-chats: bump `updated` in setAutoTitle / setUserTitle

Blacksuan19:
Title mutations did not advance the conversation's `updated` timestamp, so any logic that tracks modifications by timestamp (exports, change detection, etc.) would silently miss renames.

### DIFF
--- a/src/common/stores/chat/store-chats.ts
+++ b/src/common/stores/chat/store-chats.ts
@@ -420,12 +420,14 @@ export const useChatStore = create<ConversationsStore>()(/*devtools(*/
         _get()._editConversation(conversationId,
           {
             autoTitle,
+            updated: Date.now(),
           }),
 
       setUserTitle: (conversationId: DConversationId, userTitle: string) =>
         _get()._editConversation(conversationId,
           {
             userTitle,
+            updated: Date.now(),
             ...(!userTitle && { autoTitle: undefined }), // clear autotitle when clearing usertitle
           }),
 


enricoros:
Need to think about this - it's tricky. The 'updated' is overloaded with 2 uses: one is change tracking, as per the request purpose, the other to sort conversations on the conversations list. The reason userTitle/autoTitle do not touch the 'updated' flag, is to not "bump up" the conversation in the sorting order (i.e. renaming the title will keep the conversation where it is). Not sure what's the best approach.

Blacksuan19:
Fair point about the sort order, I haven't considered that intentional design. 

To be transparent: I've built a personal sync layer on top of Big AGI and that is what surfaced the issue, renames weren't being picked up by change detection because `updated` field wasn't being bumped.

That being said, I agree that conflating "sort recency" with "last change timestamp" in a single field is the real problem here. One clean approach would be a separate `titleUpdated?: number` field then `updated` can stay purely as the sort/activity key, while `titleUpdated` tracks title mutations. The downside is it adds a field that's currently only useful for change-tracking purposes, so I understand if that feels like over-engineering for the upstream codebase right now.

enricoros:
@Blacksuan19 Good points. How's the sync going? The separation of the variable is what I have in the free/pro version of Big-AGI on big-agi.com. So if I applied the change here it would make the downstream (the free/pro version) conflict and cost quite some work for me. There are pieces of Big-AGI that are 1:1 in sync, such as AIX (first updated in the Open and then in the free/pro), but most data-related structures I won't change here or I'll have to spend days resolving conflicts at the moment.

Blacksuan19:
Makes total sense, I hadn't considered the downstream conflict cost on data structures. Good to know the separation already exists in the hosted version; I'll just carry the titleUpdated field in my fork. Thanks for the context, closing this one.

As for sync: it's Supabase-based, works well with these changes. Conversations, blobs, and a handful of localStorage stores. Nothing fancy but does the job across devices.

enricoros:
@Blacksuan19 Nice capabilities. I can't say I endorse it, because I need the free/pro version to be successful so that I can keep developing the open-source (7,000 commits :). However very good job, I know for a fact it's not easy, and those data structures can be quite harsh.

DoS007:
@enricoros To understand? If I rename a chat and export it for a version upgrade to another deployment domain, it won't keep the renamed chat? Or is this different for "flash" exports? If so, is there a workaround, like sending a message and deleting it, after renaming?

enricoros:
@DoS007 Exporting will work correctly, in all forms. Renamed chats will not jump on the top of the list (because if one goes and cleans up old titles they don't necessarily want to have thost chats at the top, while message changes would do) but exporting has no issues, and will work correctly.

Blacksuan19:
I think chats moving to the top when renamed is the correct behavior, I only realized big agi does not do this after using librechat for a while, the list of chats should reflect the correct order of modification, why would someone rename a chat randomly if they are not going to use it? other chat frontends like chatgpt do this too

DoS007:
@enricoros Ah, ok, nice.

I probably misinterpreted the text by @Blacksuan19 (formating is mine)

>  any logic that tracks modifications by timestamp (*exports*, change detection, etc.) would silently miss renames



@Blacksuan19 Funny that you tested librechat. I also used that. For me the huge advantage of big-agi is, that it has an easy enduser export and import system for the messages and even the more general state (flash export). Although I really would like making system prompts for my "self hosted" instance. But that's not as important for me.


To topic: Tested chatgpt: chat name changes move it to the top list. But claude.ai doesn't do it. So from this sample set, there don't seem to be a de-facto standard.

_My opinion:_ Renaming should not bump chats to the top. But a raw bump/poke option would be nice. When searching through old chats because they are relevant at the moment, it might be nice to make them available together on the top for easier handling and to see what was recently relevant (easier to find more often used "old" chats). But this is only of minor convience so would consider that unimportant.

Blacksuan19:
> @enricoros Ah, ok, nice.
> 
> I probably misinterpreted the text by @Blacksuan19 (formating is mine)
> 
> > any logic that tracks modifications by timestamp (_exports_, change detection, etc.) would silently miss renames
> 
> @Blacksuan19 Funny that you tested librechat. I also used that. For me the huge advantage of big-agi is, that it has an easy enduser export and import system for the messages and even the more general state (flash export). Although I really would like making system prompts for my "self hosted" instance. But that's not as important for me.
> 
> To topic: Tested chatgpt: chat name changes move it to the top list. But claude.ai doesn't do it. So from this sample set, there don't seem to be a de-facto standard.
> 
> _My opinion:_ Renaming should not bump chats to the top. But a raw bump/poke option would be nice. When searching through old chats because they are relevant at the moment, it might be nice to make them available together on the top for easier handling and to see what was recently relevant (easier to find more often used "old" chats). But this is only of minor convience so would consider that unimportant.

I spent way too much time setting up librechat and could never get it to work near as good as Big AGI so I gave up on it and decided to try and implement sync in Big AGI myself, Big AGI is great!

I see, so there is no consensus on how this should be, I think sorting should be configurable, like based on latest message or based on chat modified time since these two things are not the same

DoS007:
@Blacksuan19 I personally wouldn't find that so important but preferences differ.

Do you also have custom system prompts? That was possible on librechat selfhosted i think.

PR #1022, opened 2026-03-08
Add Tensorix as a new LLM provider

shanemort1982:
Hey there 👋

This adds [Tensorix](https://tensorix.ai) as a new OpenAI-compatible vendor in big-AGI, letting users access models from DeepSeek, Meta Llama, Qwen, GLM, MiniMax and more through a single API key and endpoint.

Tensorix follows the same OpenAI-compatible pattern as Deepseek/Together/etc, so the integration is straightforward — new vendor folder + wiring into the existing dialect system.

### What's included

- **New vendor**: `src/modules/llms/vendors/tensorix/` — vendor definition + setup UI component
- **Dialect**: `tensorix` added to `openai.access.ts` with default host `https://api.tensorix.ai`
- **Server env**: `TENSORIX_API_KEY` for server-side deployment config
- **Backend capability**: `hasLlmTensorix` for the green checkmark when pre-configured
- **Registry + UI wiring**: vendor registered in `vendors.registry.ts`, `LLMVendorSetup.tsx`, and `backend.router.ts`

### How it works

Users add their Tensorix API key in Models → Add → Tensorix and models are fetched from the `/v1/models` endpoint. The setup UI includes an optional custom host field and client-side fetch toggle, matching the pattern of other cloud vendors.

This also opens up cross-promotion opportunities — we'd be happy to feature big-AGI prominently in Tensorix's integration docs at [docs.tensorix.ai](https://docs.tensorix.ai).

Happy to adjust anything if needed!

### DIFF
--- a/src/modules/backend/backend.router.ts
+++ b/src/modules/backend/backend.router.ts
@@ -64,6 +64,7 @@ export const backendRouter = createTRPCRouter({
         hasLlmOpenPipe: !!env.OPENPIPE_API_KEY,
         hasLlmOpenRouter: !!env.OPENROUTER_API_KEY,
         hasLlmPerplexity: !!env.PERPLEXITY_API_KEY,
+        hasLlmTensorix: !!env.TENSORIX_API_KEY,
         hasLlmTogetherAI: !!env.TOGETHERAI_API_KEY,
         hasLlmXAI: !!env.XAI_API_KEY,
         // others
--- a/src/modules/backend/store-backend-capabilities.ts
+++ b/src/modules/backend/store-backend-capabilities.ts
@@ -24,6 +24,7 @@ export interface BackendCapabilities {
   hasLlmOpenPipe: boolean;
   hasLlmOpenRouter: boolean;
   hasLlmPerplexity: boolean;
+  hasLlmTensorix: boolean;
   hasLlmTogetherAI: boolean;
   hasLlmXAI: boolean;
   // others
@@ -66,6 +67,7 @@ const useBackendCapabilitiesStore = create<BackendStore>()(
     hasLlmOpenPipe: false,
     hasLlmOpenRouter: false,
     hasLlmPerplexity: false,
+    hasLlmTensorix: false,
     hasLlmTogetherAI: false,
     hasLlmXAI: false,
     hasDB: false,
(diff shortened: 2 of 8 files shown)


DoS007:
The service seems to be similiar to avian (but maybe better)? https://github.com/enricoros/big-AGI/pull/996

Can it be, that people try to get money from that openrouter use case pie?

shanemort1982:
Hey — actually we're quite different from Avian/OpenRouter. Tensorix isn't an aggregator or proxy — we host all our models on our own hardware. Zero data retention, no third-party routing, your requests never leave our infrastructure.

So it's more like running your own inference stack but without the ops headache. We offer popular open-source models (DeepSeek, Llama, GLM, MiniMax, etc.) through a standard OpenAI-compatible API, pay-as-you-go. But the key difference is everything runs on infra we own and operate — nothing gets forwarded to another provider.

Happy to answer any other questions!

DoS007:
@shanemort1982 Ah, ok. So more like deepinfra and others.

enricoros:
Considering merge of this one if we see interest or requests by our users.

JiwaniZakir (commented):
In `openai.access.ts`, the guard condition `if (!tensorixKey || !tensorixHost)` will never trigger for the host portion, since `tensorixHost` is derived from `llmsFixupHost(access.oaiHost || DEFAULT_TENSORIX_HOST, apiPath)` — the `DEFAULT_TENSORIX_HOST` fallback ensures it's always non-empty. The error message "Missing Tensorix API Key or Host" is therefore misleading; the `!tensorixHost` branch is dead code and should be removed to keep the guard honest (just `if (!tensorixKey)`), consistent with how similar dialects like `xai` handle it.

Additionally, looking at other dialect cases in the same `switch` block, the `tensorix` case introduces a `let` declaration (`let tensorixKey`) followed by a `const` on the next line, then reassigns `tensorixKey`. This is consistent with other cases but placing the `llmsRandomKeyFromMultiKey` call directly in the initializer (`let tensorixKey = llmsRandomKeyFromMultiKey(access.oaiKey || env.TENSORIX_API_KEY || '')`) would make the mutation unnecessary and cleaner.

It's also worth confirming whether `https://api.tensorix.ai` is publicly documented as stable — the comment references `https://docs.tensorix.ai` but the linked dashboard URL in `TensorixServiceSetup.tsx` (`https://app.tensorix.ai/dashboard`) suggests this is a relatively new provider; if the base API URL is subject to change during early access, a fallback note or env-var override documentation would be worth adding to `.env.example`.

enricoros:
Thanks for the contribution - closing this: we're not adding hosted-vendor integrations without demonstrated user demand, and none has surfaced since March. Tensorix works today via the OpenAI-compatible custom-host path, so users who want it aren't blocked. Happy to revisit if requests come in.

PR #1032, opened 2026-03-15
chore: upgrade Next.js to 16 while preserving webpack direct connection

alexx-ftw:
## Summary
Upgrade the app to Next.js 16 and React 19 while preserving the existing webpack-based direct-connection path and the main Edge tRPC runtime.

## What changed
- Upgraded `next`, `react`, `react-dom`, `eslint-config-next`, and `@next/bundle-analyzer`
- Updated scripts for Next 16, including replacing `next lint` with `eslint .`
- Kept webpack explicitly enabled for `dev` and `build` so the existing custom webpack hook remains active
- Updated the ESLint flat config to the current `eslint-config-next/core-web-vitals` format
- Kept the main tRPC endpoint on Edge at `/api/edge`
- Restored the webpack client-module replacements and client mocks used by the direct-connection flow
- Kept the TypeScript config aligned with current Next output expectations

## Why webpack is still enabled
This repo still relies on the custom webpack module replacement in `next.config.ts` to swap `~/server/env.server` and `~/server/posthog/posthog.server` with client mocks in browser bundles:
- `next.config.ts:51-94`
- `src/server/env.client-mock.ts:1-28`
- `src/server/posthog/posthog.client-mock.ts:1-27`

That path is needed to preserve the direct-connection behavior referenced in review.

## Runtime behavior retained
The main tRPC endpoint remains on Edge:
- `app/api/edge/[trpc]/route.ts:17-20`

The public endpoint path is unchanged:
- `app/api/edge/[trpc]/route.ts:6-20`

## Validation
Passed:
- `npm run lint`
- `npm run build`

Build currently runs on webpack with Next 16 because the direct-connection path still depends on the custom webpack hook.

## Notes
- This PR no longer includes the previous Turbopack migration or Edge-to-Node runtime migration attempt.
- The existing Edge runtime build warning remains expected while `app/api/edge/[trpc]` stays on Edge.


enricoros:
This is directionally good, but also removes the much used "direct connection" feature for AI models. Besides that (which is a stopper for sure), I will have to verify bundles sizes composition, to maybe get parts of this feature in.

A couple of comments:
- upgrade of React should be good
- upgrade of Next: should be good, besides checking bundles sizes and composition, and that we don't add server-side rendering. Directionally, I'd love to depend on Vercel less, not more, because of the runtime (5min) and size (4MB) limitations for the requests
- edge removal: leaning towards no, because right now the edge pops are routing the AI requests to the servers closer to the user, while with the node runtime, they'll go to 1-3 servers globally; this will increase latency and I'm not sure what the upside would be
- web pack: as mentioned this will make "direct connection" stop working for AI models, so unless there's an equivalent, I can't get this part in

alexx-ftw:
Thanks, I’ve trimmed this PR accordingly. It now keeps the Edge runtime, restores the webpack path and client mocks needed for direct connection, and keeps the Next 16 / React 19 / ESLint upgrade pieces. I also revalidated with npm run lint and npm run build.

enricoros:
@alexx-ftw Thanks! I'll review manually, comparing the bundles contents before/after and checking network paths. If things check out (also on the pro/free branches I maintain for big-agi.com) I'll merge.

enricoros:
Hi @alexx-ftw I've tested this, but the change is not great for the product: the update to React 19 adds to the client bundle size and the update to Next 16 reduces the visibility (e.g. stats after build) and increases opacity and dependence on Vercel. At the same time, this doesn't give new features that we'd use or will make people's life easier (and as mentioned, I would love to reduce the dependency on Vercel, as right now has runtime and size and architecture limitations that are damaging to the product). So for now I'll close this one, and we can revisit in the future if we need some of the new React19/Next16 features.

alexx-ftw:
Of course man no problem at all. As long as Next 15 keeps being maintained staying is totally fine.
FYI 
image
 That is EOL for 15

PR #1033, opened 2026-03-16
feat(ui): add system theme mode for dark mode controls

Blacksuan19:
## Summary

This PR adds system theme support to the app's dark mode controls.

The theme provider now defaults to Joy UI's `system` mode, and the existing dark mode toggle now cycles through `Light Theme`, `Dark Theme`, and `System Theme` instead of only forcing a light/dark switch.

## Why

Previously, users had to manually toggle the app theme when their desktop or browser theme changed. With this change, big-AGI can follow the operating system theme automatically while still allowing manual overrides.

## Changes

- set the Joy `CssVarsProvider` default mode to `system`
- update the theme toggle UI to support `light`, `dark`, and `system`
- refresh the button labels, icons, and tooltip text to reflect the active theme preference

## Validation

Tested manually on KDE Plasma by switching the system color scheme between light and dark variants. The app followed the system theme correctly in both a regular browser window and VS Code Simple Browser.

### DIFF
--- a/src/common/components/DarkModeToggleButton.tsx
+++ b/src/common/components/DarkModeToggleButton.tsx
@@ -1,8 +1,10 @@
 import * as React from 'react';
 
 import { Button, IconButton, useColorScheme } from '@mui/joy';
+import BrightnessAutoIcon from '@mui/icons-material/BrightnessAuto';
 import DarkModeIcon from '@mui/icons-material/DarkMode';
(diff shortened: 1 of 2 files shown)


enricoros:
@Blacksuan19 Thanks for the change - great you're using Plasma :) (I was a long time KDE contrib)

Blacksuan19:
> I was a long time KDE contrib

Oh wow. small world 😀

PR #1069, opened 2026-04-18
Keep beams

dagelf:
This change makes completed Beam outputs durable without polluting the conversation history.

Before this, Beam only kept its intermediate/completed ray and fusion results in the in-memory Beam store. A reload lost them, and they were absent from exports/backups unless the user explicitly accepted one back into chat as a normal assistant message.

Now, completed Beam outputs are persisted as conversation side-data in conversation.beamResults, and that field is included in the existing conversation import/export path. That means:
  - completed Beam results survive reloads
  - they are included in backups/exports
  - they are still excluded from normal chat context because they are not added to conversation.messages

Runtime behavior is intentionally narrow:
  - only completed rays and completed fusion outputs are saved
  - in-flight Beam execution state is not persisted
  - accepting a Beam result into chat still behaves exactly as before and writes a normal assistant message into history
  - reopening Beam seeds the Beam UI from saved beamResults using the existing importRays path
  
  I used codex with gpt-5.4.


dagelf:
https://github.com/enricoros/big-AGI/issues/1068

dagelf:
Works for me. There might be permutations I have not tried, and I'm not sure if it will affect other setups. 
I've tried reloading, exporting, importing, 

Closing beams with `esc` still wipes them. But that's the only way to wipe them now.

enricoros:
@dagelf Great patch with low lines of code, those are the best.

Do you also use the big-agi.com product, or are you comfortable using your fork here?

GitHub vs big-agi.com have different features set (common AI pipeline and most UI tho), in particular big-agi.com has a more sophisticated persistence scheduler that requires a different shape of the patch. 

An official solution will need me to patch it on both places but also enable a few things that are related in the roadmap: 1. standalone Beams, 2. per-message Beams (so we can go back to a specific message and change our mind and pick another Beam variant), 3. Beams of other parameters such as permutations of Personas, and more. 

If you're comfortable using your branch until then, then this patch is well done and the way to go. Just let me know if you need this on big-agi.com anytime soon as well.

dagelf:
You can try the CI, there's a good chance it won't break main, and at least beams will be recoverable from backups.

enricoros:
Note, this feature has not been landed yet.

enricoros:
Leaving open, not adopting.

The goal is right and it's planned: durable Beam outputs are coming as a first-class synced entity (beam sessions) tied into persistence, sync, and import/export - a system, not side-data on the conversation. That's also why this shape can't merge: conversation.beamResults changes the conversation schema, which the hosted branch syncs with a different persistence scheduler, and the two would conflict on every rebase until sessions land.

The feature arrives, differently. Keeping this open as the record of what was asked for.

PR #1156, opened 2026-07-07
T2I: OpenRouter: add image generation via the dedicated Image API

Robinnnnn:
## Motivation

OpenRouter recently launched a dedicated image generation endpoint (`POST /api/v1/images` — [docs](https://openrouter.ai/docs/features/multimodal/image-generation)). This lets big-AGI users generate images with the many image models available on OpenRouter (Gemini Image / Nano Banana, GPT Image, FLUX.2, Seedream, Recraft, Riverflow, MAI, ...) through a single OpenRouter key — the same one already configured for chat. Today big-AGI has OpenAI/Azure/LocalAI T2I providers, but no way to draw through OpenRouter.

## Implementation

Follows the existing T2I provider pattern (DALL·E is the reference):

- **Server**: new `dialectOpenRouter_createImages` streaming procedure in `openai.router.ts` (heartbeats while awaiting, same as `createImages`), posting `{ model, prompt, n? }` to `/v1/images` via the existing `openrouter` access dialect (so `OPENROUTER_API_KEY` server env and client-configured keys both work). Response wiretype (`{ created, data: [{ b64_json, media_type? }], usage? }`) added to `openrouter.wiretypes.ts`. Since `media_type` can be absent (and is sometimes inaccurate), the mime type and dimensions are sniffed from the decoded bytes with the existing `getImageInformationFromBytes` helper, falling back to `media_type`/png for formats it doesn't know (e.g. WebP).
- **Client**: new `src/modules/t2i/openrouter/` module — generation client, a small persisted store for the selected image model, and a settings panel with a starter model list (the full catalog is linked). OpenRouter services are picked up by `_findLlmsT2IServices` / `_getTextToImageProviders`, and the `openrouter` vendor is wired into `t2iGenerateImagesOrThrow`.
- **UI**: model selector appears in Draw > Options and in Preferences > Draw; the Draw service selector shows the OpenRouter icon.

Not included (kept minimal): image-to-image (the endpoint supports it via `images[]`, but wiring the edit path is a natural follow-up), per-model size/quality options, dynamic model listing from `/api/v1/models?output_modalities=image`.

## Testing

- `tsc --noEmit` and `eslint` clean on all touched files.
- Manually tested end-to-end against the live OpenRouter API with a real key: generated images through the Draw app UI (service selector → prompt → Draw), verified the image renders in the Gallery with correct origin metadata (`generatorName: google/gemini-2.5-flash-image`, 1024x1024, `image/png` sniffed from bytes, input/output token usage populated), and verified abort/error paths return readable messages.
- Also exercised the tRPC route directly (batched jsonl streaming) to confirm the heartbeat + `createImage` op flow.

Happy to adjust the starter model list, naming, or move the model selection elsewhere — opening as draft for direction.

### DIFF
--- a/src/apps/draw/create/DrawProviderConfigure.tsx
+++ b/src/apps/draw/create/DrawProviderConfigure.tsx
@@ -5,6 +5,7 @@ import { Box, Button, Card, CardContent } from '@mui/joy';
 import ConstructionIcon from '@mui/icons-material/Construction';
 
 import { DallESettings } from '~/modules/t2i/dalle/DallESettings';
+import { OpenRouterT2ISettings } from '~/modules/t2i/openrouter/OpenRouterT2ISettings';
 
 import type { TextToImageProvider } from '~/common/components/useCapabilities';
 import { ExpanderControlledBox } from '~/common/components/ExpanderControlledBox';
@@ -29,7 +30,7 @@ export function DrawProviderConfigure(props: {
 
   const { ProviderConfig } = React.useMemo(() => {
     const provider = providers.find(provider => provider.providerId === activeProviderId);
-    const ProviderConfig: React.FC | null = provider?.vendor === 'openai' ? DallESettings : null;
+    const ProviderConfig: React.FC | null = provider?.vendor === 'openai' ? DallESettings : provider?.vendor === 'openrouter' ? OpenRouterT2ISettings : null;
     return {
       ProviderConfig,
     };
--- a/src/apps/draw/create/DrawProviderSelector.tsx
+++ b/src/apps/draw/create/DrawProviderSelector.tsx
@@ -7,6 +7,7 @@ import KeyboardArrowDownIcon from '@mui/icons-material/KeyboardArrowDown';
 
 import type { TextToImageProvider } from '~/common/components/useCapabilities';
 import { OpenAIIcon } from '~/common/components/icons/vendors/OpenAIIcon';
+import { OpenRouterIcon } from '~/common/components/icons/vendors/OpenRouterIcon';
 import { hideOnMobile } from '~/common/app.theme';
 import { optimaSelectSlotProps } from '~/common/layout/optima/bar/OptimaBarDropdown';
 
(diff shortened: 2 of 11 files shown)


Robinnnnn:
@enricoros would appreciate a review when you have a chance! 🙏

Moving this out of draft. Adds OpenRouter as a T2I provider via the dedicated `POST /api/v1/images` endpoint, following the existing DALL·E pattern (server streaming procedure with heartbeats + wiretype, new `src/modules/t2i/openrouter/` client module, Draw/Preferences UI wiring).

Did another self-review pass and made two adjustments in the latest commit:
- **Multi-image fan-out.** Most OpenRouter image models cap `n=1` (all Gemini/FLUX/MAI/Riverflow in the starter list), so sending `n=count` would silently return a single image for those. Now fans out `count` parallel single-image requests — same approach the DALL·E 3 path uses — so the requested count is honored regardless of a model's `n` cap. Uses the same `Promise.allSettled` + AbortError-preserving error handling as the DALL·E client.
- **Logging consistency.** Switched the non-fatal image-sniff failure from `console.log` to `console.warn`, matching the `console.error`/`warn` convention elsewhere in `openai.router.ts`.

`tsc --noEmit` and `eslint` are clean on all touched files. Still intentionally minimal — image-to-image, per-model size/quality, and dynamic model listing are called out as follow-ups. Happy to adjust the starter model list or move the model selector.

enricoros:
@Robinnnnn Thanks! I haven't had the time to manually verify this yet.

@claude How does this work, is it optimal, does it require maintenance or porting efforts? Also with the t2i system being a bit behind speex/voice (as far as configurability, auto detection etc) does this support auto detection well?

enricoros:
@claude perfect..now read the OpenRouter image Gen API docs to understand whether there's a good coverage of the features there.

enricoros:
@claude fair. Merge able? Potential conflict with evolution of t2i more like speex/asrx?

enricoros:
@Robinnnnn Merged - and extensive framework and UI changes around the configuration; will easily support any other providers in the future.

PR #1167, opened 2026-08-04
Bedrock: support Mantle models that only implement the OpenAI Responses API

worsnoptr:
## What

Adds a `mantle-responses` value to the `llmVndBedrockAPI` model parameter, dispatching Bedrock Mantle models through the OpenAI **Responses** API instead of Chat Completions, plus a small fix so `LLM_IF_HOTFIX_NoTemperature` survives the Bedrock model listing for Anthropic models.

## Why

Some Bedrock Mantle models only implement the OpenAI Responses API. Today that's the OpenAI frontier family (GPT-5.4, GPT-5.5, GPT-5.6 Sol / Terra / Luna): they appear in the model list (via Mantle `GET /v1/models`) but every invoke fails with:

> **400**: The model 'openai.gpt-5.6-terra' does not support the '/v1/chat/completions' API

because the `mantle` dispatch hardcodes `POST /v1/chat/completions`.

Note the path detail: per the AWS model cards, these models are served on the **`/openai/v1/responses`** path — different from the `/v1/responses` path used by other models on the responses endpoint (see the "API compatibility by models" page and e.g. the GPT-5.5 model card in the AWS Bedrock user guide).

## How

- `llms.parameters.ts` / `aix.wiretypes.ts`: extend the `llmVndBedrockAPI` enum with `mantle-responses`.
- `chatGenerate.dispatch.ts`: new `case 'mantle-responses'` — `bedrockURLMantle(region, '/openai/v1/responses')`, body from the existing `aixToOpenAIResponses` adapter, parsed with `createOpenAIResponsesEventParser('openai')` / `createOpenAIResponseParserNS('openai')`, signed with the existing `bedrockAccessAsync` (no auth changes).
- `bedrock.models.ts`: the `KNOWN_MANTLE_ONLY` map gains an `api: 'responses'` marker (matched exactly or with a trailing `-YYYY-MM-DD` snapshot suffix stripped, e.g. `openai.gpt-5.4-2026-03-05`), with metadata per the AWS docs (272K context, vision, reasoning). These models also reject `temperature` ("Unsupported parameter: 'temperature' is not supported with this model."), so they carry `LLM_IF_HOTFIX_NoTemperature` like their OpenAI-vendor counterparts.
- `anthropic.models.ts` (second commit): `_BEDROCK_ANT_IF_ALLOWLIST` was stripping `LLM_IF_HOTFIX_NoTemperature` from Claude models listed via Bedrock, so clients kept sending `temperature` and Bedrock rejected it with `400: "\`temperature\` is deprecated for this model."` (Claude Sonnet 5 base / Fable 5). The direct Anthropic vendor was unaffected. The flag is now allowlisted.

## Tested

All verified live against `bedrock-mantle.us-east-1.api.aws` (SigV4):

- `openai.gpt-5.6-terra` via `mantle-responses`: streaming and non-streaming ✅
- No regressions on the other three Bedrock paths: `zai.glm-4.7-flash` (mantle / Chat Completions), Nova Pro (converse), Claude Sonnet 5 + Claude Fable 5 (invoke-anthropic) ✅
- Model listing: all Responses-only IDs (including dated snapshots) get the `mantle-responses` spec; gpt-oss / Grok models that support both APIs stay on `mantle` ✅
- `temperature` regression: Bedrock-listed Sonnet 5 / Fable 5 and the GPT-5.x Mantle models now carry `hotfix-no-temperature`, so clients omit temperature and the 400s are gone ✅
- `tsc --noEmit` clean, `next build` passes

## Also included: server-configured vendors not auto-appearing for new users

Two related fixes for backend-configured services (found while deploying a Bedrock-only instance):

- **`reconfigureBackendModels.ts`**: the reconfig hash was persisted *before* the configuration loop, so a transient failure in `llmsUpdateModelsForServiceOrThrow` (swallowed by the catch) permanently skipped auto-configuration for that browser. The hash is now persisted after the loop, and only when at least one service configured successfully (or there was nothing to configure) — on total failure the next boot retries. Resolves the existing `FIXME` on that line; the in-session double-run guard is unchanged.
- **`backend.router.ts`**: `generateLlmEnvConfigHash` filtered env keys with `key.includes('_API_')`, which excludes all `BEDROCK_*` vars — Bedrock-only deployments never (re-)triggered client autoconfiguration. The filter now also matches `key.startsWith('BEDROCK_')`. The env values only feed a one-way hash, same as the existing `*_API_KEY` values.

### DIFF
--- a/src/common/logic/reconfigureBackendModels.ts
+++ b/src/common/logic/reconfigureBackendModels.ts
@@ -34,8 +34,6 @@ export async function reconfigureBackendModels(lastLlmReconfigHash: string, setL
 
   // begin configuration
   _isConfiguring = true;
-  // FIXME: future: move this to the end of the function, but also with strong retry count and error catching, so one's app wouldn't loop upon each boot
-  setLastReconfigHash(backendReconfigHash);
   const initiallyEmpty = !llmsStoreState().llms?.length;
 
   // reconfigure these
@@ -64,6 +62,7 @@ export async function reconfigureBackendModels(lastLlmReconfigHash: string, setL
 
   // track in order the services that were configured
   const configuredServiceIds: DModelsServiceId[] = [];
+  let configurationSuccesses = 0;
 
   // sequentially re-configure
   await servicesToReconfigure.reduce(async (promiseChain, service) => {
@@ -74,6 +73,7 @@ export async function reconfigureBackendModels(lastLlmReconfigHash: string, setL
 
         // auto-configure this service
         await llmsUpdateModelsForServiceOrThrow(service.id, true);
+        configurationSuccesses++;
       })
       .catch(error => {
         // catches errors and logs them, but does not stop the chain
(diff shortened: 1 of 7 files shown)


enricoros:
@worsnoptr Good job here - have been reviewing this` - candidate for merge today

PR #1168, opened 2026-08-04
Bedrock: opt-in ambient IAM via the container credentials endpoint

worsnoptr:
## What

Opt-in ambient AWS IAM credentials for Bedrock via the **container credentials endpoint** (ECS/Fargate task roles): set `BEDROCK_USE_CONTAINER_CREDENTIALS=true` and big-AGI authenticates with the task's IAM role — no static `BEDROCK_ACCESS_KEY_ID` / `BEDROCK_SECRET_ACCESS_KEY` needed.

## Why

Enterprises with IAM-role-only security mandates cannot use long-lived static credentials — a hard requirement in many orgs (see e.g. [BerriAI/litellm#29665](https://github.com/BerriAI/litellm/issues/29665), which documents deployments where "a long-lived Bedrock API key ... is a non-starter under the security posture: no long-lived static secrets, IAM-role-only access"; litellm has a trail of similar asks). For big-AGI on ECS/Fargate this currently forces an IAM user + secret distribution + rotation machinery that the platform already solves natively with task roles.

## How — Edge Runtime compatible, zero dependencies

The existing constraint ("SigV4 uses explicit credentials only, no SDK credential chain, for Edge Runtime compatibility") is honored: **no AWS SDK is introduced**. On ECS the agent injects `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI`, and a plain `fetch` of `http://169.254.170.2<uri>` returns `{ AccessKeyId, SecretAccessKey, Token, Expiration }` as JSON. The existing `aws4fetch` signer already accepts `sessionToken`, so short-lived role credentials need **no signing changes**.

Empirical Edge-sandbox verification (self-hosted `next start`, which runs `runtime = 'edge'` routes in the edge-runtime sandbox): `globalThis.EdgeRuntime === 'edge-runtime'`, no Node `process.version`, yet `process.env` is readable and `fetch` reaches the credentials endpoint (measured <100 ms; it's link-local). On Vercel/Cloudflare the env var never exists, so the code path stays dormant.

- **`bedrock.containerCredentials.ts`** (new, ~110 lines): endpoint detection (`AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` / `_FULL_URI`, optional static `AWS_CONTAINER_AUTHORIZATION_TOKEN`), module-level cache, refresh 5 minutes before `Expiration`, single-flight concurrent refreshes, stale-tolerant on refresh failure while credentials remain valid, one-time log of the assumed role. Out of scope by design (documented in the header): EC2 IMDSv2, EKS IRSA, EKS Pod Identity's rotating token *file* (needs `fs`, unavailable on Edge).
- **`bedrock.access.ts`**: ambient credentials slot in at the **lowest priority** — client bearer > client IAM > server bearer > server IAM env > container credentials (opt-in). Explicit configuration always wins; existing deployments see zero behavior change unless they set the flag. `_bedrockResolveAuth` becomes async; `bedrockResolveRegion` is split out to stay sync and credential-free (same region semantics as before).
- **`backend.router.ts`**: `hasLlmBedrock` also lights up when the flag is set and the endpoint is present, so client auto-configuration works for keyless deployments.
- **`env.server.ts`** + **`docs/environment-variables.md`**: the new variable.

## Tested

- **Live on ECS Fargate (us-east-1), task role only** (no credential env vars in the task definition): model listing (140 models) plus chats across all Bedrock API paths — Converse (Nova Pro), Invoke-Anthropic (Claude Sonnet 5, Fable 5), and Mantle. Container logs show the one-time `[Bedrock] using container credentials (role: arn:aws:iam::…:role/…)` line.
- **Credential lifecycle**: a single credentials fetch served an entire session of concurrent listing + chat requests (cache + single-flight verified).
- **Priority**: with explicit env credentials present alongside the flag, the ambient endpoint is never contacted.
- **Failure path**: with the flag set but no endpoint (or fetch failure) and no other credentials, the standard "Missing AWS credentials" error is returned and a `console.warn` explains the ambient failure; a still-valid cached credential keeps serving through transient refresh failures.
- `tsc --noEmit` clean, `next build` passes.

## Notes

- Pairs well with (but does not depend on) the Bedrock Responses-API PR from the same deployment work; both are independently mergeable from `main`.
- Infra simplification for deployers: an ECS task role statement replaces the IAM user, secret distribution, and rotation machinery entirely; container env shrinks to `BEDROCK_REGION` + `BEDROCK_USE_CONTAINER_CREDENTIALS=true`.

### DIFF
--- a/docs/environment-variables.md
+++ b/docs/environment-variables.md
@@ -34,6 +34,7 @@ BEDROCK_ACCESS_KEY_ID=
 BEDROCK_SECRET_ACCESS_KEY=
 BEDROCK_SESSION_TOKEN=
 BEDROCK_REGION=
+BEDROCK_USE_CONTAINER_CREDENTIALS=
 DEEPSEEK_API_KEY=
 GEMINI_API_KEY=
 GROQ_API_KEY=
@@ -109,6 +110,7 @@ requiring the user to enter an API key
 | `BEDROCK_SECRET_ACCESS_KEY` | AWS IAM Secret Access Key for Bedrock                                                                          | Optional, but if set `BEDROCK_ACCESS_KEY_ID` must also be set      |
 | `BEDROCK_SESSION_TOKEN`     | AWS Session Token for temporary/STS credentials                                                                | Optional                                                          |
 | `BEDROCK_REGION`            | AWS region for Bedrock (e.g., `us-east-1`, `us-west-2`, `eu-west-1`)                                          | Optional, defaults to `us-east-1`                                 |
+| `BEDROCK_USE_CONTAINER_CREDENTIALS` | Set to `true` to use ambient IAM credentials from the ECS/Fargate container credentials endpoint (task role) when no explicit Bedrock credentials are configured. Lowest priority in the credentials chain | Optional                                                          |
 | `DEEPSEEK_API_KEY`          | The API key for Deepseek AI                                                                                    | Optional                                                          |
 | `GEMINI_API_KEY`            | The API key for Google AI's Gemini                                                                             | Optional                                                          |
 | `GROQ_API_KEY`              | The API key for Groq Cloud                                                                                     | Optional                                                          |
--- a/src/modules/backend/backend.router.ts
+++ b/src/modules/backend/backend.router.ts
@@ -6,6 +6,8 @@ import { createTRPCRouter, publicProcedure } from '~/server/trpc/trpc.server';
 import { env } from '~/server/env.server';
 import { fetchJsonOrTRPCThrow } from '~/server/trpc/trpc.router.fetchers';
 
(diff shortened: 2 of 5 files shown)


enricoros:
Leaving open, not merging.

The craft is high - pure-fetch, Edge-compatible, the credential lifecycle handling is right. The math is what fails: this serves ECS/Fargate task-role deployments, a thin slice of self-hosters, and costs permanent surface (auth-chain priority, capability wiring, docs) on files the hosted branch reworks on every rebase. Narrow benefit, standing cost. Deployments that need it can carry the patch - it applies cleanly.

If the audience is bigger than I think, make the case and I'll reconsider.

PR #1170, opened 2026-08-04
WebSearch: add Brave Search as a server-configured provider

worsnoptr:
## Why — Google is sunsetting whole-web Programmable Search Engines

Per the Programmable Search Engine blog (Jan 2026), creating new **whole-web** Programmable Search Engines has been blocked since January 2026, and existing whole-web engines must transition off by **2027-01-01**. That is exactly the configuration big-AGI's web search uses: every deployment with `GOOGLE_CLOUD_API_KEY` + `GOOGLE_CSE_ID` doing whole-web search (the `/react` flow's `search` action) loses its provider within the year. Domain-restricted engines survive, but the general "search the web" use case needs a replacement path.

## What

Adds the [Brave Search API](https://brave.com/search/api/) (`api.search.brave.com/res/v1/web/search`, `X-Subscription-Token` auth) as a server-configured web search provider behind the existing abstraction — one new env var, zero client protocol changes.

- **`BRAVE_SEARCH_API_KEY`** server env var (documented in `docs/environment-variables.md`) + new **`hasBraveSearch`** backend capability.
- The `googleSearch.search` tRPC procedure selects the provider server-side and keeps returning the same provider-neutral `BriefResult[]` pages — the `/react` flow and all client code are untouched.
- **Selection order**: client-provided Google keys keep priority (explicit user configuration), then server-side Brave, then server-side Google CSE. Brave is deliberately preferred over server Google since Google's product is being sunset.
- **Fidelity**: domain restriction maps to Brave's `site:` query operator; Brave's highlight markup (`<strong>`) and HTML entities are stripped so snippets remain plain text, matching Google's output shape.
- **Settings UI**: when the server has Brave configured, the search settings treat search as backend-fulfilled (Google fields become optional overrides, with an explanatory note).

## Behavior guarantees

- **Unset = dormant**: without `BRAVE_SEARCH_API_KEY`, nothing changes — verified that capabilities and the search endpoint behave byte-for-byte as before (including the exact same error when no provider is configured).
- Existing Google-only deployments are unaffected; users with their own Google keys in the UI keep using them even if the server has Brave.

## Tested (live against the Brave API)

- Whole-web queries return correct pages with clean plain-text snippets ✅
- `restrictToDomain` returns only results from the requested domain (via `site:`) ✅
- Client-provided Google keys still route to Google (override verified) ✅
- Both server providers configured → Brave is used ✅
- No providers configured → stock error, zero behavior change ✅
- `tsc --noEmit` clean, `next build` passes

## Follow-up candidates

Tavily (and similar search APIs) can slot into the same server-side selection with the same `BriefResult` mapping; this PR deliberately implements Brave only.

### DIFF
--- a/docs/environment-variables.md
+++ b/docs/environment-variables.md
@@ -55,6 +55,7 @@ PUPPETEER_WSS_ENDPOINT=
 # Search
 GOOGLE_CLOUD_API_KEY=
 GOOGLE_CSE_ID=
+BRAVE_SEARCH_API_KEY=
 
 # Text-To-Speech: ElevenLabs
 ELEVENLABS_API_KEY=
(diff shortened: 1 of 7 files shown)


enricoros:
Not merging as-is, leaving open.

The motivation is real - whole-web PSE sunsets 2027-01-01 and that is exactly the config /react search uses - and the implementation is clean. What blocks it is direction, not craft: search is getting a provider framework rather than per-provider forks inside search.router - multiple implementors (Brave, Jina, Exa, vendor-native search where a model has it, MCP-based), user-configurable, exposed as web/fetch tools across the app instead of only the /react loop. Same pattern as our multi-vendor TTS abstraction. Your Brave mapping is exactly an implementor for that framework, and this PR is where I'll come take it from.

Leaving open until the framework lands.

Issue #1173, opened 2026-08-04
[Feature] Per-conversation model binding — multi-turn, multi-model Multi-Chat


worsnoptr:
# [Feature] Per-conversation model binding — multi-turn, multi-model Multi-Chat

## The gap

Multi-Chat split view with broadcast (#388) already does the hard part: one message goes to N conversations, each with its own history. But every pane executes on the same app-level `primaryChat` model assignment — so side-by-side comparison of *different models over a multi-turn conversation* isn't possible today.

**This is not Beam.** Beam is single-turn scatter/gather with merge — brilliant for "give me 4 takes on this one prompt and fuse them". What's missing is the longitudinal version: *persistent conversations, each bound to its own model, advancing together turn by turn* — watching how Claude vs GPT vs Nova handle the same evolving dialogue, context drift and all. Beam and this compose rather than compete: you'd still Beam within any single pane (and in the sketch below, Beam seeds from the pane's bound model).

## Proposed design: `conversation.userLlmId`

An **optional** pinned model on the conversation, resolved at execution time:

```
effective model = validated(conversation.userLlmId) ?? primaryChat domain assignment
```

- `DConversation.userLlmId?: DLLMId` — absent on every existing chat = follow the default. No `app-chats` version bump; export carries it; import round-trips it (one-line add in `chats.converters` `recreateConversation`); `duplicateDConversation` copies it so branches keep their model.
- **Broken pins degrade silently** to the domain default (model deleted, or chat imported from an install with different services) — same semantics as broken domain assignments in `llmsResolveDomainModel`. No migration, no pruning pass.
- The resolution swap lives at the existing choke point: `_handleExecute` reads `getConversationChatLLMId(conversationId)` instead of `getChatLLMId()`. Because the multicast loop already dispatches per conversation, **broadcast with per-pane models needs zero changes to the broadcast code**. Beam's seed (`beamOpen`/`beamImportRays` in `ConversationHandler`) resolves per-conversation the same way.
- **Utility flows are untouched**: autotitle/fastUtil/codeApply/imageCaption use `getDomainModelIdOrThrow` — separate accessors by design.

## UX (conservative — zero behavior change until you pin)

- **Model selector** (top bar, operates on the focused pane's conversation): unchanged by default — selecting a model still sets the app-level default. New menu entry: **"Pin model to this chat"** / **"Unpin · follow app default"**. When pinned: the selector shows a 📌 on the active model and edits the pin instead of the global default.
[...]

PR #1174, opened 2026-08-04
Chat: per-conversation model binding for multi-model Multi-Chat

worsnoptr:
Implements the design proposed in #1173 — see that issue for the full rationale, UX mock, and discussion of the pane-binding alternative. Opening as a draft so the design conversation in the issue can proceed independently of code review; happy to adjust (field name, UX details, or switch to pane-binding) based on how that discussion goes.

## Summary

- `DConversation.userLlmId?: DLLMId` — optional pinned model, resolved at execution time as `pin ?? primaryChat domain default`.
- `getConversationChatLLMId(conversationId)`: validated pin or default; broken pins (model removed, or chat imported from an install with different services) silently degrade to the default, mirroring `llmsResolveDomainModel`'s broken-assignment semantics.
- `_handleExecute` reads it instead of `getChatLLMId()` — one swap covers send, regenerate, retry, ReAct, and Multi-Chat broadcast (which already dispatches per conversation, so per-pane models fall out with no changes to the broadcast code itself). Beam's seed (`beamOpen`/`beamImportRays`) resolves per-conversation too.
- Utility flows (autotitle, fastUtil, codeApply, imageCaption) are untouched — they use the separate domain accessors.
- UX: model selector gains a "Pin model to this chat" / "Unpin · follow app default" entry; selecting a model edits the pin when pinned, the app default otherwise (zero behavior change until you pin). Each pane's title overlay now shows its bound model — outlined chip + 📌 when pinned, dimmed italic when following the default — so non-focused panes' models stay visible for side-by-side comparison.
- Persistence: optional field, no store version bump; `duplicateDConversation` copies the pin so branches keep their model; export carries it and import round-trips it (`chats.converters`).

## Tested

- Single-pane, no pin: byte-identical behavior to before (verified via the send/regenerate/ReAct paths reading the same resolved model as `getChatLLMId()` would return).
- Pin / unpin / re-pin via the selector menu; broken-pin degradation (delete the pinned service, confirm fallback to default).
- Duplicate/branch carries the pin; export → import round-trip.
- 4-pane broadcast with four different models across all Bedrock dispatch paths (Converse, Invoke-Anthropic, Mantle Chat Completions, Mantle Responses) — verified each pane continues multi-turn on its own model and the per-model tool state (e.g. a Responses-only model's Web Search toggle) stays scoped to its own pane.
- `tsc --noEmit`, `eslint`, and `next build` all clean.

### DIFF
--- a/src/apps/chat/AppChat.tsx
+++ b/src/apps/chat/AppChat.tsx
@@ -28,8 +28,7 @@ import { addSnackbar, removeSnackbar } from '~/common/components/snackbar/useSna
 import { createDMessageFromFragments, createDMessagePlaceholderIncomplete, DMessageMetadata, duplicateDMessageMetadata } from '~/common/stores/chat/chat.message';
 import { createErrorContentFragment, createTextContentFragment, DMessageAttachmentFragment, DMessageContentFragment, duplicateDMessageFragments } from '~/common/stores/chat/chat.fragments';
 import { gcChatImageAssets } from '~/common/stores/chat/chat.gc';
-import { getChatLLMId } from '~/common/stores/llms/store-llms';
-import { getConversation, getConversationSystemPurposeId, useConversation } from '~/common/stores/chat/store-chats';
+import { getConversation, getConversationChatLLMId, getConversationSystemPurposeId, useConversation } from '~/common/stores/chat/store-chats';
 import { optimaActions, optimaOpenModels, optimaOpenPreferences, useOptimaChromeless } from '~/common/layout/optima/useOptima';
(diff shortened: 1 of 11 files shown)


enricoros:
Thanks for opening this as a draft - DConversation.userLlmId is the right first move. To make it land cleanly: (1) keep the field; (2) route all chat model resolution through one function (chat pin > persona default > global fallback) instead of reading the pin at call sites - that mirrors how our hosted branch already works, and it is the difference between merging cleanly and fighting every rebase; (3) degrade silently when a pinned model disappears (fall back, never error); (4) leave the utility model domains (title/fast) alone. Reshape to that and review follows quickly.

PR #1178, opened 2026-08-18
ReAct: include recent chat history in the agent context

githubespresso407:
## Summary

Implements the long-standing TODO in `react.ts` (`to initialize with previous chat messages to provide context`): the ReAct agent now receives a compact transcript of the recent conversation, so follow-up questions like "what about the second one?" resolve against the actual chat subject instead of failing or hallucinating.

## How

- `react-tangent.ts` builds a transcript from the last 6 user/assistant messages (600 chars each), skipping the `/react` command message itself and the freshly appended assistant placeholder.
- The transcript is prepended as quoted context inside the question rather than injected as role messages into the loop's buffer, so the `Action:` regex parser can't trip on content from earlier turns.
- No history (first message of a conversation) → byte-identical behavior to today.

## Testing

- `tsc --noEmit` clean on current `main`.
- Manual: two-turn chat, then `/react` follow-up with a deictic reference → the agent searches for the right subject and answers.

### DIFF
--- a/src/apps/chat/editors/react-tangent.ts
+++ b/src/apps/chat/editors/react-tangent.ts
@@ -3,10 +3,38 @@ import { useBrowseStore } from '~/modules/browse/store-module-browsing';
 
 import type { ConversationHandler } from '~/common/chat-overlay/ConversationHandler';
 import type { DLLMId } from '~/common/stores/llms/llms.types';
-import { createErrorContentFragment, createTextContentFragment } from '~/common/stores/chat/chat.fragments';
+import type { DMessage } from '~/common/stores/chat/chat.message';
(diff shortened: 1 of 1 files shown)


enricoros:
Leaving open, not merging.

Right diagnosis, and the guardrails (quoted context rather than role messages, so the Action: parser can't trip) are thought through. The reason to hold: the ReAct runner is being replaced - the hosted branch already runs it on an execution engine that feeds full chat history, and that engine is headed to main. Merging this creates a conflict on that exact file for every rebase between now and then, for behavior the migration delivers anyway.

If the migration slips far, I'll revisit - this is a fine stopgap.

Issue #209, opened 2023-11-16
Add Confetti


enricoros:
Seems like the perfect reward after achieving some milestones, such as:
 - completing a first > 1min call
 - generating the first diagram
 - adding one of the local model providers
 - accepting a flattened conversation
 - new version of the app is released, with major changes (/news page)

Requirements:
 - https://www.npmjs.com/package/react-canvas-confetti
 - use this lib, seems small footprint
 - confetti need to have the same color as the app theme
 - when not used, the canvas has to Not be there, at all, disappeared, gone
 - bonus if the library is dynamically loaded

Fun feature!

edmondop:
Can I pick this one?

enricoros:
> Can I pick this one?

For sure!

enricoros:
Closing as not planned - reversing my 2023 'for sure'. A confetti library did land on our hosted branch as a stray dependency and it is on its way out; we are not adding another one to main (PR #1185, declined with this). A third-party package for a cosmetic effect is supply-chain surface we do not want. The one celebration moment that earns it can be hand-rolled if it ever matters.

PR #1185, opened 2026-08-26
Add Confetti celebration for first diagram generated

tatianajrogel:
## Summary
Implements a first, focused piece of #209 (Add Confetti): fires a theme-colored confetti burst the first time a user successfully generates a diagram, then fully unmounts it.

## What changed
- `ConfettiTrigger.tsx`: lazy-loads `react-canvas-confetti` so the canvas/confetti bundle stays out of the main chunk. The `<canvas>` is fully unmounted (not just hidden) once the burst finishes, per the issue's efficiency requirement.
- `store-celebrations.ts`: a small persisted zustand store (same shape as the existing `store-client.ts`) tracking which one-off milestones have already been celebrated, so a given milestone only ever fires once per device.
- Wired into `DiagramsModal.tsx`'s existing generation-success path.
- Colors come from the Joy UI theme CSS vars (`--joy-palette-primary-500` / `--joy-palette-neutral-500`) with static fallbacks, so it matches the active theme instead of hardcoded colors.

## Why only one milestone
The issue lists several trigger points (first >1min call, new release, first diagram, adding a local model provider, accepting a flattened conversation). Rather than half-wire all of them, I picked one (first diagram generated) and implemented it end-to-end, verified. `celebrationConsumeMilestone()` is written so the other triggers can reuse the same store in follow-up PRs — happy to add more if that's wanted here instead.

## Verified locally
- `npm ci` (matches CI's install step) succeeds with the updated `package-lock.json`.
- `npx tsc --noEmit` — no errors in the new/changed files.
- `npm run lint` — clean.

Towards #209

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -32,6 +32,7 @@
         "@vercel/speed-insights": "^2.0.0",
         "aws4fetch": "^1.0.20",
         "browser-fs-access": "^0.38.0",
+        "canvas-confetti": "^1.9.3",
         "cheerio": "^1.1.2",
         "csv-stringify": "^6.8.3",
         "dexie": "~4.0.11",
@@ -49,6 +50,7 @@
         "prismjs": "^1.30.0",
         "puppeteer-core": "^24",
         "react": "^18",
+        "react-canvas-confetti": "^2.0.7",
         "react-dom": "^18",
         "react-hook-form": "^7.85.0",
         "react-markdown": "^10.1.0",
@@ -68,6 +70,7 @@
       },
       "devDependencies": {
         "@posthog/nextjs-config": "~1.6",
+        "@types/canvas-confetti": "^1.9.0",
         "@types/node": "^22",
         "@types/nprogress": "^0.2.3",
         "@types/prismjs": "^1.26.6",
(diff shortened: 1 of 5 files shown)


tatianajrogel:
Hi @enricoros — this is my first PR to this project. Could someone approve the CI workflow run so it can execute? I've verified `npm ci`, `tsc --noEmit`, and `npm run lint` all pass locally, so hopefully CI is straightforward. Happy to iterate on any feedback. Thanks!

tatianajrogel:
Hi @enricoros — first PR to big-AGI. The CI run is gated on maintainer approval (first-time contributor). Could you approve the workflow so it can run? The change adds a lightweight, lazy-loaded confetti celebration component (see #209) with no impact on the main bundle. Verified locally with `npm ci`, `tsc --noEmit`, and `npm run lint`. Thanks!

enricoros:
Thanks for the clean implementation - the lazy-load and the full unmount after the burst are the right instincts, genuinely. Declining anyway: this adds a runtime dependency (react-canvas-confetti) for a cosmetic effect, and we are tightening the dependency surface, not growing it - a confetti library that slipped onto our hosted branch as a stray is on its way out, not becoming a precedent. Supply-chain cost is not justified here. Closing #209 with the same rationale, so this is a policy call, not a review of your code.

enricoros:
@tatianajrogel Appreciated this - however this was for an older issue (confetti) which does not justify the added dependencies, especially given we really need to be aware of "supply chain attack" risks (when an upstream library like confetti is compromised).
If you're looking for an issue to contribute on, find your own personal pain or let down with the application, and I'll accept patches on that.

Issue #221, opened 2023-11-17
[BUG] Voice input doesn't stay on


ramicaza:
**Describe the bug**
Absolutely love the voice recognition feature! But in a previous version, I would be able to speak, collect my thoughts, then keep speaking. Now whenever there's a pause for a few seconds to think, it ends speech recognition, extremely annoying!  I think this bug was introduced when the 'Voice continuation' feature was added which makes sense because when that mode is enabled, it should automatically determine when to stop 'listening' and send the message. But I'd wager many users (including myself) rarely use 'voice continuation'

Where is it happening?
- All platforms

**To Reproduce**
Use the 'voice input' with voice continuation disabled, say a few things, stop, then see it stop recording.

**Expected behavior**
See above

ramicaza:
I am on Mac. Your proposal makes sense, but if I'm being honest I don't think I would use it because I would likely just switch voice continuation off. I can think of times when I'd take much longer than 5s to collect my thoughts when speaking to it. (maybe i'm just slow 🤠) 

I believe with voice continuation off, it should never stop transcribing unless the user hits the microphone button to explicitly tell it to stop listening. With voice continuation on, your proposal makes perfect sense. To me, I am mainly concerned about the voice continuation off case.

enricoros:
Thanks, I'll prototype this (maybe with a 30s max delay or an "off" option).

The complexity is handling 2 different logics (mobile and desktop behave differently) in all 3 modes (mic is also used for the "Call" function, where I need a timeout signal).

Good inputs.

enricoros:
Added a configurable timeout which works well both for people who want to have quick automatic chats, as well as people that can want to collect thoughts, and something in between.

![image](https://github.com/enricoros/big-agi/assets/32999/0874dd8c-a160-4ac0-acfa-d126f2b3a860)

What do you think?

ramicaza:
Love it!!
Just one minor bug is that you need to refresh the page for the change to take effect.

enricoros:
Fixed it, thanks for reporting @ramicaza!

PR #1189, opened 2026-08-28
Call: respect the shared Voice Input Mic Timeout

### LINKED ISSUE #1188: Call: configurable pause before reply (end-of-turn silence)
## Why

Speech recognition in Call mode may stop while the user is still speaking.

Changing Voice -> Input -> Mic Timeout does not affect Call mode. Even when the timeout is set to 15 seconds, Call may stop listening shortly after receiving an interim recognition result.

## Description

Chat Composer reads the shared `micTimeoutMs` setting and passes it to `useSpeechRecognition()`.

Call uses the same speech recognition hook, but `Telephone.tsx` currently passes a hardcoded timeout of `1000` ms:

```ts
useSpeechRecognition('webSpeechApi', onSpeechResultCallback, 1000);
```

The timeout is restarted after each Web Speech API `onresult` event. Therefore, if the browser does not emit another result within one second, Call stops recognition even if the user is still speaking.

As a result, the Voice -> Input -> Mic Timeout setting applies to Chat but is ignored by Call.

## Requirements

- Call must use the shared Voice Input Mic Timeout setting.
- Call and Chat must use the same timeout behavior and fallback value.
- Changing Mic Timeout must update the active speech recognition configuration.
- Existing Push to Talk and continuous listening behavior must remain unchanged.
- This change must not alter the timeout implementation inside `WebSpeechApiEngine`.

## Implementation

- Read `micTimeoutMs` in `Telephone.tsx` using `useChatMicTimeoutMsValue()`.
- Replace the hardcoded `1000` ms value with `chatMicTimeoutMs || 2000` (like a Chat Composer).
- Keep the implementation consistent with Chat Composer.

## Verification

1. Open Voice -> Input -> Mic Timeout and select `15s`.
2. Start a Call.
3. Enable Push to Talk.
4. Speak continuously, including short pauses.
5. Confirm that Call no longer stops after the previous hardcoded one-second deadline.
6. Change the setting to another value and confirm that Call uses the updated timeout.

## Contribute
I would like to contribute a fix. I have tested a local implementation of this change and can submit a Pull Request. 

### Where does this happen?

Self-deployed from GitHub

### Impact on your workflow

High - Major feature broken

### Environment (if applicable)

Environment: Docker container
Device: MacBook pro M1
OS: MacOs 13.7.2
Browser: Chrome Version 151.0.7922.174

### Additional context

_No response_

SergeyAblaev:
The change is minimal, matches the Composer pattern, and preserves Push-to-Talk / continuous behavior.

## Why

Call mode ignores the Voice Input Mic Timeout setting and uses a hardcoded one-second timeout. This may stop speech recognition while the user is still speaking.

## Changes

- Read the shared `micTimeoutMs` setting in `Telephone`.
- Pass the configured timeout to `useSpeechRecognition`.
- Match the fallback behavior used by Chat Composer.

## Testing

- IntelliJ project build passes.
- `git diff --check` passes.
- Manually tested Call with Mic Timeout set to 2,5,15 seconds.

Fixes #1188

### DIFF
--- a/src/apps/call/Telephone.tsx
+++ b/src/apps/call/Telephone.tsx
@@ -11,6 +11,7 @@ import MicNoneIcon from '@mui/icons-material/MicNone';
 import { ScrollToBottom } from '~/common/scroll-to-bottom/ScrollToBottom';
 import { ScrollToBottomButton } from '~/common/scroll-to-bottom/ScrollToBottomButton';
 import { useChatLLMDropdown } from '../chat/components/layout-bar/useLLMDropdown';
+import { useChatMicTimeoutMsValue } from '../chat/store-app-chat';
 
 import { SystemPurposeId, SystemPurposes } from '../../data';
 
@@ -89,6 +90,7 @@ export function Telephone(props: {
   const responseAbortController = React.useRef<AbortController | null>(null);
 
   // external state
+  const chatMicTimeoutMs = useChatMicTimeoutMsValue();
   const { chatLLMId: modelId, chatLLMDropdown: modelDropdown } = useChatLLMDropdown(llmDropdownRef);
   const { chatTitle, reMessages } = useChatStore(useShallow(state => {
     const conversation = props.callIntent.conversationId
@@ -114,7 +116,7 @@ export function Telephone(props: {
         setCallMessages(messages => [...messages, createDMessageTextContent('user', userSpeechTranscribed)]); // [state] append user:speech
     }
   }, []);
-  const { recognitionState, startRecognition, stopRecognition, toggleRecognition } = useSpeechRecognition('webSpeechApi', onSpeechResultCallback, 1000);
+  const { recognitionState, startRecognition, stopRecognition, toggleRecognition } = useSpeechRecognition('webSpeechApi', onSpeechResultCallback, chatMicTimeoutMs || 2000);
 
   // derived state
   const isRinging = stage === 'ring';


enricoros:
Thanks for the PR, and sorry for the mixed signal - the auto-triage comment green-lit this, but I'm not taking it in this form.

The 1000 in Telephone isn't the chat mic timeout applied badly, it's the call's end-of-turn detector: when it fires, your sentence is sent and the persona answers. The chat setting has a different job (when dictation stops), which is why its default is 5s and it goes up to 15s. Wiring the two together makes every call wait 5 seconds of silence before the reply by default, and 15 for people who set "Best for thinking" for dictation. I kept them separate on purpose in #221.

If you want a longer pause, I'd take it as a call-specific option instead: a value in `src/apps/call/state/store-app-call.ts` (default 1000) and a "Pause before reply" item in the Call menu next to Push to talk, 1s / 2s / 3s, passed to `useSpeechRecognition` in Telephone. Same size as this PR, no coupling to chat. Want to rework #1189 into that?

SergeyAblaev:
@enricoros  Okay, I got the idea. Yes, one second isn't enough for a thoughtful conversation or language practice with the AI.
Okay, I'll try adding a separate "Mic Timeout for call" setting next to the chat setting. Options like 1, 2, 3, 5 seconds - I think that would be more convenient, right?

enricoros:
Landed a broader version of this on main in 2b760a6e07: a call-local "Send after" control in the call panel (1s / 2s / 5s / 15s, default 2s) plus a "Settings" mode that follows the chat Mic Timeout, which is what this PR wanted, now opt-in instead of default. Closing in favor of that; thanks for the diagnosis and the push on #1188, both credited there.

enricoros:
@SergeyAblaev Thanks for your help here - I had requests for this feature on a separate thread and had to implement this (and ship on main) today.

